PR #35, opened 2019-02-27
fix: App does not crash after pressing send button multiple number of times.

### LINKED ISSUE #32: IllegalArgumentException when SEND button click 3 times
**Describe the bug**

App Stopped working due to FATAL EXCEPTION: main.

**To Reproduce**

Install App and allow Bluetooth, with no device connected. With empty field, click on SEND more than 3 times.

**Expected behavior**

If no device is connected, the device should give a message, may be in the form of Toast that device not connected or Scan could not find any device.

**Logs**

E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.nilhcem.blenamebadge, PID: 10565
    java.lang.IllegalArgumentException: scanner already started with given callback
        at no.nordicsemi.android.support.v18.scanner.BluetoothLeScannerImplLollipop.startScanInternal(BluetoothLeScannerImplLollipop.java:61)
        at no.nordicsemi.android.support.v18.scanner.BluetoothLeScannerCompat.startScan(BluetoothLeScannerCompat.java:112)
        at com.nilhcem.blenamebadge.device.bluetooth.ScanHelper.startLeScan(ScanHelper.kt:57)
        at com.nilhcem.blenamebadge.ui.message.MessagePresenter.sendBytes(MessagePresenter.kt:36)
        at com.nilhcem.blenamebadge.ui.message.MessagePresenter.sendBitmap(MessagePresenter.kt:25)
        at com.nilhcem.blenamebadge.ui.message.MessageActivity$onCreate$3.onClick(MessageActivity.kt:58)
        at android.view.View.performClick(View.java:6614)
        at android.view.View.performClickInternal(View.java:6587)
        at android.view.View.access$3100(View.java:784)
        at android.view.View$PerformClick.run(View.java:26125)
        at android.os.Handler.handleCallback(Handler.java:873)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:193)
        at android.app.ActivityThread.main(ActivityThread.java:6715)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:911)
I/Process: Sending signal. PID: 10565 SIG: 9
Application terminated.

**Screenshots**

image

**Smartphone Info:**
|                            |                       |
|-------------------|----------------|
|Device                 |Nokia 5.1 Plus |
|Android Version  |Android Pie     |

Yes

yashk2000:
Fixes #32 

Changes: App does not crash after pressing send button multiple number of times when there is no device connected. A toast is shown.

Screenshots for the change:

screencap

### DIFF
--- a/app/src/main/java/com/nilhcem/blenamebadge/ui/message/MessageActivity.kt
+++ b/app/src/main/java/com/nilhcem/blenamebadge/ui/message/MessageActivity.kt
@@ -27,10 +27,13 @@ import com.nilhcem.blenamebadge.device.model.DataToSend
 import com.nilhcem.blenamebadge.device.model.Message
 import com.nilhcem.blenamebadge.device.model.Mode
 import com.nilhcem.blenamebadge.device.model.Speed
+import java.util.Timer
+import java.util.TimerTask
 
 class MessageActivity : AppCompatActivity() {
 
     companion object {
+        private const val SCAN_TIMEOUT_MS = 10_000L
         private const val REQUEST_ENABLE_BT = 1
         private const val REQUEST_PERMISSION_LOCATION = 1
     }
@@ -54,6 +57,13 @@ class MessageActivity : AppCompatActivity() {
 
         send.setOnClickListener {
             // Easter egg
+            send.isClickable = false
+            val buttonTimer = Timer()
+            buttonTimer.schedule(object : TimerTask() {
+                override fun run() {
+                    runOnUiThread { send.isClickable = true }
+                }
+            }, SCAN_TIMEOUT_MS)
             if (content.text.isEmpty()) {
                 presenter.sendBitmap(this, BitmapFactory.decodeResource(resources, R.drawable.mix2))
             } else {


yashk2000:
please review @mariobehling

app/src/main/java/com/nilhcem/blenamebadge/ui/message/MessageActivity.kt
```diff
@@ -54,10 +54,14 @@ class MessageActivity : AppCompatActivity() {
...
-                presenter.sendBitmap(this, BitmapFactory.decodeResource(resources, R.drawable.mix2))
-            } else {
-                presenter.sendMessage(this, convertToDeviceDataModel())
+            try {
+                if (content.text.isEmpty()) {
+                    presenter.sendBitmap(this, BitmapFactory.decodeResource(resources, R.drawable.mix2))
+                } else {
+                    presenter.sendMessage(this, convertToDeviceDataModel())
+                }
+            } catch (e: Exception) {
```
iamareebjamal: Please do not catch generic exceptions

yashk2000: @iamareebjamal could please tell me what else to do instead?

yashk2000:
@iamareebjamal the app crashes if send button is pressed multiple times while the app is scanning for devices after the send button was clicked the first time. Could we fix this issue by making the send button non-clickable for the time interval when the scan is going on, i.e. `private val SCAN_TIMEOUT_MS = 10_000L`?

yashk2000:
@iamareebjamal I set a timer on the button to be disabled for the time period when the scan is already going on. Now the button is not clickable during the time and hence the app does not crash even if the button is pressed. Should I update this pull request with the changes?

iamareebjamal:
Timeout shouldn't be hard-coded. Enable the button once the scan is done

yashk2000:
@iamareebjamal I have hardcoded the time out because the time for scanning devices is already hardcoded in the ScanHelper class. The time is declared implicitly in this class and a postDelayed handler is used to control the duration of the scan. 
That's why I thought it was alright to declare the time limit variable in MessageActivity and use it in the timer. Should I keep it this way only or should it be changed?

CloudyPadmal:
Hi, with this implementation, the button will always be blocked for `SCAN_TIMEOUT_MS` right? As @iamareebjamal mentioned, can you make an implementation to enable the button once the scan is complete?

yashk2000:
@CloudyPadmal once the scan is over, the button gets activated again. The button is first deactivated, then using the timer, I activate the button after the scan time is over. I tested it thoroughly on my device, it works.

CloudyPadmal:
@yashk2000 what if the scan takes longer than 10 seconds? What if a user device is slower than your device and it takes more time to scan? In a scenario like this, the best thing to do is have a flag or something to check if the scan is over and then activate the button.

yashk2000:
@CloudyPadmal I do think that the scan will never take more than 10 seconds. In the ScanHelper class, where the scan starts, a post delayed handler is used to control the time. This is the code used in the ScanHelper class:
```kotlin
fun startLeScan(onDeviceFoundCallback: ((BluetoothDevice?) -> Unit)) {
        this.onDeviceFoundCallback = onDeviceFoundCallback
        isScanning = true

        val filters = listOf(ScanFilter.Builder()
                .setServiceUuid(ParcelUuid(SERVICE_UUID))
                .build())

        val settings = ScanSettings.Builder()
                .setScanMode(ScanSettings.SCAN_MODE_LOW_LATENCY)
                .build()

        scanner.startScan(filters, settings, scanCallback)

        // Stops scanning after a pre-defined scan period.
        stopScanHandler.postDelayed(stopScanRunnable, SCAN_TIMEOUT_MS)
    }
```

`stopScanHandler.postDelayed(stopScanRunnable, SCAN_TIMEOUT_MS)` will prevent the scan from running for more than 10 seonds. Even if the other device is slow, 10 seconds is the time for which the scan will run no matter what. I have used the same thing to enable and disable the button so I think that duration of both the scan and disabling the button will be same, irrespective of the device.

Please tell me if this is correct or not. If it is not, I try using a flag to check when the scan is over and enable the button.

yashk2000:
@iamareebjamal @CloudyPadmal please review and tell me if I have to make any changes.

mariobehling:
Ok, let's go with this and see if this is the way as we move on. Would be great, if you could help to implement more features as well. Thank you.

yashk2000:
@mariobehling I would really love to work on developing the app further, but I do not have the hardware device needed. Is there any way to work without the hardware?

mariobehling:
Ok, great. There is a link to order it from China on the readme. We will also be able to sell it through the PSLab.io website in a few days, but ordering from China is cheaper.

yashk2000:
@mariobehling I will be attending the FOSSASIA tech summit in Singapore, from 14-17 March. Is there any way I could get the hardware device directly in Singapore?

yashk2000:
> @mariobehling I will be attending the FOSSASIA tech summit in Singapore, from 14-17 March. Is there any way I could get the hardware device directly in Singapore while I'll be there for the summit?

@mariobehling could you please answer this question? It would be really great if I could get the hardware directly when I come to Singapore next week.

Issue #14, opened 2018-12-16
Implement a badge preview on the top of the screen


mariobehling:
There are a number of apps to use the phone as an LED screen. Please take inspiration and implement a preview screen to show how it will look on the actual device on the top. Compare for example to this app: https://play.google.com/store/apps/details?id=com.rooftopSpace.Led

dilum1995:
Hello @mariobehling ,

Is it possible for me to work on this and also to improve the application UI?

mariobehling:
Please go ahead. As this project is still early stage there is no need to ask if you can take over an issue. Simple get started according to your interests. Thank you.

PR #53, opened 2019-03-12
Badge Preview Top Screen

anshulmak:
Part of for issues #14 

Fixed a part of the issue#14 by implementing a badge preview at the top of the screen and syncing the message to be send with the text of badge preview.

Screenshots for the change:

Screenrecorder201903121859


yashk2000:
I don't think this is what is needed. We need a preview of how it looks in the actual badge. This is simply a representation of the text the user types in a black background.

anshulmak:
> I don't think this is what is needed. We need a preview of how it looks in the actual badge. This is simply a representation of the text the user types in a black background.

anshulmak:
> I don't think this is what is needed. We need a preview of how it looks in the actual badge. This is simply a representation of the text the user types in a black background.

I have used an led screen image as background just like an actual badge and after this i was thinking to apply the animations to the text according to the mode selected in the app. 

Can you please recommend me what can i do to implement a badge preview screen?

sparrow007:
> I don't think this is what is needed. We need a preview of how it looks in the actual badge. This is simply a representation of the text the user types in a black background.

@yashk2000 would you show how it will looks for clarity ??

PR #63, opened 2019-03-21
Disable send button when no device connected

### LINKED ISSUE #60: Disable Send button when there is no bluetooth device connected
A message should show up when there is no bluetooth device connected. For example a toast message or an alert.

**Additional context**
We could also disable the send button when there is no bluetooth device connected.

Yes

yatri1609:
Fixes #60 

Changes: 
Made Send button disabled or non-clikable when there is no bluetooth device connected.

Screenshots for the change:
Screenrecorder-2019-03-21-10-43-57-198

### DIFF
--- a/app/src/main/java/com/nilhcem/blenamebadge/ui/message/MessageActivity.kt
+++ b/app/src/main/java/com/nilhcem/blenamebadge/ui/message/MessageActivity.kt
@@ -4,6 +4,7 @@ import android.Manifest
 import android.app.Activity
 import android.app.AlertDialog
 import android.bluetooth.BluetoothAdapter
+import android.bluetooth.BluetoothDevice
 import android.bluetooth.BluetoothManager
 import android.content.Context
 import android.content.Intent
@@ -79,6 +80,7 @@ class MessageActivity : AppCompatActivity() {
             }
         }
         prepareForScan()
+        isDeviceConnected()
     }
 
     override fun onResume() {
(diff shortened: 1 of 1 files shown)


mariobehling:
In this case, the color should change e.g. the text of the send button more dark gray. Also a message should be displayed on a pop up "No device connected."

app/src/main/java/com/nilhcem/blenamebadge/ui/message/MessageActivity.kt
```diff
@@ -151,10 +153,22 @@ class MessageActivity : AppCompatActivity() {
             }
         } else {
             Toast.makeText(this, "BLE is not supported", Toast.LENGTH_LONG).show()
+            send.isClickable = false
```
iamareebjamal: That's not correct logic. You should disable the button instead. Making it non clickable will only annoy the users who won't even know why is it not registering their clicks. Also, as Mario said, there should be an error message saying that there is no device connected

yatri1609: Yeah got that I would do the changes

yashk2000:
@iamareebjamal @mariobehling this is not needed. The app does not require a Bluetooth device to be connected. It uses BLE to send the message. If we check for a Bluetooth connection and disable the button, the button will always be disabled. When the send button js clicked, if the badge is in the range, the message is sent using BLE, but the badge is never connected using Bluetooth. It is also not visible in Bluetooth devices in settings menu.

PR #78, opened 2019-03-22
fix: Button disabled instead of being non-clickable.

### LINKED ISSUE #74: Send button becomes non clickable after pressing once for 10s.
Right now, the send button is non-clickable after being pressed once for 10s. Instead it should just be disabled so that even if a user presses it, it appears to be clickable.

YES

yashk2000:
Fixes #74 

Changes: The send button is now disabled for 10s after being clicked and is then enabled again instead of being non-clickable.

GIF for the change:

ezgif com-video-to-gif (15)


yashk2000:
@liveHarshit The send button becomes a faded grey color to show that it is disabled. And when the button is enabled again, it turns back to the normal color.

yashk2000:
@iamareebjamal please review.

PR #102, opened 2019-03-30
fix: Added missing parameters in setValue Function

adityastic:
Changes:
- Added the rest of parameters while calling setValue() in MessageActivity


adityastic:
@iamareebjamal The last two merged PR had compilation issues because the parameters were changed for the function in one of the PR. Please have a look at this

PR #107, opened 2019-03-30
fix: Created a seperate file for landscape orientation.

### LINKED ISSUE #106: Visibility of badge in landscape mode.
**Describe the bug**
In landscape mode, the badge preview is not fully visible.

**To Reproduce**
Shift to landscape mode.

**Expected behavior**
Badge preview should be completely visible.

**Screenshots**

screencap

**Additional Context**
The size of the text "Badge Magic" also decreases on the top. That should also not happen.

YES.

yashk2000:
Fixes #106 

Changes: A new layout resource file has been added for landscape orientation.

Screenshots for the change:

screencap

### DIFF
--- a/app/src/main/res/layout-land/message_activity.xml
+++ b/app/src/main/res/layout-land/message_activity.xml
@@ -1,0 +1,186 @@
+<?xml version = "1.0" encoding="utf-8"?>
+<ScrollView xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
+
+    <LinearLayout
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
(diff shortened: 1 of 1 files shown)


yashk2000:
@iamareebjamal @adityastic @yatri1609 please review.

iamareebjamal:
No need to recreate entire layout just for a widget. Either, fix the widget or use dimens

yashk2000:
@iamareebjamal the height of the preview badge is fixed to 135dp in the xml. In landscape mode, the same height remains, while the height has to be increased so that the complete badge is visible. It is not a problem with the widget. 

> No need to recreate entire layout just for a widget. Either, fix the widget or use dimens

Could you please tell me what do you mean by use dimens?

iamareebjamal:
create a dimen and override in values-land

iamareebjamal:
@yashk2000 It should not be fixed then right? It should maintain its aspect ratio like all other widgets

PR #108, opened 2019-03-30
fix: Fix height of preview badge in landscape mode.

yashk2000:
Fixes #106 
Changes: Fixed the height of the preview badge in landscape mode.

Screenshots for the change:

image

PR #111, opened 2019-03-31
fix: the preview widget now maintains it's ratio according to the resolution

adityastic:
Fixes #108 

Changes: 
- added onMeasure to always match the ratio provided


adityastic:
@yashk2000 @yatri1609 @iamareebjamal. Review the Changes

Issue #132, opened 2019-04-03
Flash is not working with left/right/up/down animation


anhanh11001:
**Describe the bug**
Flash is not working with left/right/up/down animation, currently, it is only working with fixed

**Screenshots**
<img src="https://i.ibb.co/Jyk3KW6/56310267-2406842126016950-5889206779149025280-n.png" width="300">

**Would you like to work on the issue?**
Yes

adityastic:
Yes it has not been implemented. Go on

PR #137, opened 2019-04-03
fix: Enable flash on left/right/up/down animation

anhanh11001:
Details:
- Remove long redundant code on setting MODE
- Set the boolean logic for flash so that other animation can works with flash. The boolean flashLEDOn means that when that flag is set to true, the LED works normally, otherwise, all LED is turned off.This flag is only changed when user set pick the ifFlag option

Fixes: #132

Screenshots for the change:
This is the screenshot for RIGHT, other MODE I have tested and it also works fine. The GIF is a little bit unclear on blinking because I think it is already be compressed and some of the frame is cut out.


anhanh11001:
@yashk2000 @yatri1609 @naman653 Please peer review. By the way, I'm starting to this project so please if there are any tips on working with this, I would love to here. Thanks guys

adityastic:
Good Job.

Issue #129, opened 2019-04-03
App crash on clicking send twice


anhanh11001:
**Describe the bug**
When user click send twice the app crashes

**Screenshots**
<img src="https://i.ibb.co/b7YHmkg/ezgif-2-cc6544cfac77.gif" width="300">

**Smartphone Info:**
Huawei Honor 7, API 24

**Would you like to work on the issue?**
Yes

PR #138, opened 2019-04-03
fix: app crash on clicking send while scanning

anhanh11001:
Detail:
Only enable startScan function if isScanning flag is false

Fixes: #129

Screenshots for the change:

### DIFF
--- a/app/src/main/java/com/nilhcem/blenamebadge/device/bluetooth/ScanHelper.kt
+++ b/app/src/main/java/com/nilhcem/blenamebadge/device/bluetooth/ScanHelper.kt
@@ -43,23 +43,22 @@ class ScanHelper {
     }
 
     fun startLeScan(onDeviceFoundCallback: ((BluetoothDevice?) -> Unit)) {
-        if (!isScanning) {
-            this.onDeviceFoundCallback = onDeviceFoundCallback
-            isScanning = true
+        if (!isScanning) return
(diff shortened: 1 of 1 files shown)


anhanh11001:
@yashk2000 @yatri1609 @naman653 Please peer review. By the way, I'm starting to this project so please if there are any tips on working with this, I would love to here. Thanks guys

yashk2000:
Invalid pull request and issue. The app is supposed to send a default bitmap to the device if the edittext is empty.

anhanh11001:
@yashk2000 I think you misunderstand the PR and the issue. It's not about empty edittext, it's about clicking send twice, which invoke the function startLeScan from ScanHelper twice and crash the device

app/src/main/java/com/nilhcem/blenamebadge/device/bluetooth/ScanHelper.kt
```diff
@@ -43,21 +43,23 @@ class ScanHelper {
     }
 
     fun startLeScan(onDeviceFoundCallback: ((BluetoothDevice?) -> Unit)) {
-        this.onDeviceFoundCallback = onDeviceFoundCallback
-        isScanning = true
+        if (!isScanning) {
```
iamareebjamal: Use early return

anhanh11001: @iamareebjamal I have fixed that

PR #142, opened 2019-04-03
fix: App crash when edittext is empty

### LINKED ISSUE #140: App crashing if edittext is empty.
**Describe the bug**
If edit text is empty, the app is crashing on pressing send.

**Expected behavior**
On pressing send when the edittext is empty, the default bitmap should be sent to the device.

YES.

yashk2000:
Fixes #140 

Changes: If edittext is empty, the loader is still switched on the default bitmap is sent to the device,


iamareebjamal:
If you fixed `App does not crash when edittext is empty`, it means you have made it to crash now

Issue #144, opened 2019-04-03
Animation running left/right/up/down doesn't go over the screen


anhanh11001:
**Describe the bug**
When choosing animation on left/right/up/down, the drawable/words don't go all the way but only go half of the screen and then stop

**Screenshots**
<img src="https://i.ibb.co/1qrz42q/ezgif-4-3527443b7b51.gif" width="300">

**Would you like to work on the issue?**
Yes

adityastic:
Yes sir, this might be great. Good luck

PR #146, opened 2019-04-04
fix: make animation on right/left/up/down goes all the way

anhanh11001:
Detail:
- Currently, valueAnimator is set to default value 8800 so if a text (Ex: length = 80) is longer than the badgeWidth (44) then only the first 44 unit of the text is accessed when calling onDraw(). Change configValueAnimation by the length of the text/drawable so that when some text is longer than the screen, the onDraw() function will loop all the length of the text

- Remove redundant code

- Set up logic to make Move (UP,DOWN,LEFT,RIGHT) goes all of the screen. Flash and Marquee still works fine

Fixes: #144

Screenshots for the change:

### DIFF
--- a/app/src/main/java/com/nilhcem/blenamebadge/ui/badge_preview/PreviewBadge.kt
+++ b/app/src/main/java/com/nilhcem/blenamebadge/ui/badge_preview/PreviewBadge.kt
@@ -141,14 +141,15 @@ class PreviewBadge : View {
                     }
                 }
 
+                val checkListLength = checkList[i].list.size
+                val checkListHeight = checkList.size
                 when (badgeMode) {
                     Mode.LEFT -> {
                         val animationValue = animationIndex.div(200)
                         if (validMarquee || flashLEDOn &&
                                 i < checkList.size &&
                                 j < checkList[i].list.size &&
-                                j >= animationValue &&
-                                checkList[i].list[j - animationValue]) {
+                                checkList[i].list[(checkListLength + j - animationValue).rem(checkListLength)]) {
                             ledEnabled.bounds = cells[i].list[j]
                             ledEnabled.draw(canvas)
                         } else {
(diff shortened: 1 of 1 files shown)


anhanh11001:
A little bit more details:
```
if (allHex.size > 0) {
            diff = ((badgeWidth - (oneByte * allHex.size)) / 2)
            if (oneByte * allHex.size < badgeWidth) {
                for (i in 0 until badgeHeight) {
                    for (j in 0 until diff) {
                        checkList[i].list.add(false)
                    }
                }
            }
        } else {
            diff = 22
            for (i in 0 until badgeHeight) {
                for (j in 0 until diff) {
                    checkList[i].list.add(false)
                }
            }
        }
```
If allHex.size is not larger than 0 than it must be equals to 0. For allHex.size = 0: 
diff = ((badgeWidth - (oneByte * allHex.size)) / 2) = badgeWidth/2 = 22 -> so it is redundant to write the lower part. I deleted this in my code

anhanh11001:
A little bit more details:
function configValueAnimation() also take argument to find the number to count down. I did this because currently, the LED light way is set to on/off (redraw) in onDraw() is based on the iteration of valueAnimator number, which set to 8800 by default. 

For example, if I set a long text, the checkList.list length will be more than the badgeWidth, let's say 2 times longer. So if I don't change the valueAnimator, it can only iterate over the first half of the long text -> half of the text is displayed all over again

anhanh11001:
@yashk2000 @yatri1609 @naman653 @adityastic guys please take a look and review. Thanks

adityastic:
Yo, Good Morning. Let me check the changed files, give me a minute

app/src/main/java/com/nilhcem/blenamebadge/ui/badge_preview/PreviewBadge.kt
```diff
@@ -141,14 +141,13 @@ class PreviewBadge : View {
...
                 when (badgeMode) {
                     Mode.LEFT -> {
                         val animationValue = animationIndex.div(200)
-                        if (validMarquee || flashLEDOn &&
-                                i < checkList.size &&
-                                j < checkList[i].list.size &&
-                                j >= animationValue &&
-                                checkList[i].list[j - animationValue]) {
+                        if (validMarquee || flashLEDOn && i < checkList.size && j < checkList[i].list.size &&
+                                checkList[i].list[(checkListLength + j - animationValue).rem(checkListLength)]) {
```
iamareebjamal: It was clearer before, arrange as it already was

anhanh11001: I have done that

PR #157, opened 2019-04-05
fix: remove code smells from app and bug fixes

### LINKED ISSUE #154: Remove code smells from app
- [ ] Remove any use of `Any`. This is not a dynamic language project, there is no need to return `Any` type from anywhere
- [ ] Use List instead of ArrayList. NEVER return or accept implementation details/concrete classes, use abstractions
- [ ] Remove casts wherever used. There is not a single instance where code written by us will use a cast. As with any use of cast by now in this project, it is just a mistake on the part of developer to first return Any or ArrayList and then cast to appropriate types

adityastic:
Fixes #154

Changes: 
- Removed Any which ultimately removed the extra casting
- All Arraylist are changed to mutableList
- convertImageToLEDHex now does not check for the black value, any value other than transparent is considered black, this removes the restriction of only grayscaled images

### DIFF
--- a/app/src/main/java/com/nilhcem/blenamebadge/adapter/DrawableAdapter.kt
+++ b/app/src/main/java/com/nilhcem/blenamebadge/adapter/DrawableAdapter.kt
@@ -41,8 +41,8 @@ class DrawableAdapter(private val context: Context, private val list: List<Drawa
         }
     }
 
-    fun getSelectedItem(): Any {
-        return if (selectedPosition == -1) -1 else list[selectedPosition]
+    fun getSelectedItem(): DrawableInfo? {
+        return if (selectedPosition == -1) null else list[selectedPosition]
     }
 
     override fun getItemCount(): Int {
--- a/app/src/main/java/com/nilhcem/blenamebadge/ui/message/MessageActivity.kt
+++ b/app/src/main/java/com/nilhcem/blenamebadge/ui/message/MessageActivity.kt
@@ -122,9 +122,10 @@ class MessageActivity : AppCompatActivity() {
         }
 
         previewButtonDrawable.setOnClickListener {
-            if (drawableRecyclerAdapter.getSelectedItem() != -1)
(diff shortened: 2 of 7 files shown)


adityastic:
@yashk2000 @yatri1609 @iamareebjamal Please Review

app/src/main/java/com/nilhcem/blenamebadge/adapter/DrawableAdapter.kt
```diff
@@ -41,8 +41,8 @@ class DrawableAdapter(private val context: Context, private val list: List<Drawa
         }
     }
 
-    fun getSelectedItem(): Any {
-        return if (selectedPosition == -1) -1 else list[selectedPosition]
+    fun getSelectedItem(): DrawableInfo? {
+        return if (selectedPosition == -1) null else list[selectedPosition]
```
iamareebjamal: Don't return null, either return default drawable or throw exception

app/src/main/java/com/nilhcem/blenamebadge/ui/message/MessageActivity.kt
```diff
@@ -142,7 +143,7 @@ class MessageActivity : AppCompatActivity() {
     private fun setupRecycler() {
         drawableRecyclerView.layoutManager = LinearLayoutManager(this, LinearLayoutManager.HORIZONTAL, false)
 
-        val listOfDrawables = ArrayList<DrawableInfo>()
+        val listOfDrawables = mutableListOf<DrawableInfo>()
```
iamareebjamal: Use listOf

adityastic:
@iamareebjamal Please have a look sir

app/src/main/java/com/nilhcem/blenamebadge/ui/message/MessageActivity.kt
adityastic: done

PR #174, opened 2019-04-07
feat: app can be used without bluetooth

### LINKED ISSUE #172: Preview without bluetooth permission
A user should be able to preview the badge even if the Bluetooth is disabled.

Yes

naman653:
Fixes #172 

Changes: App can be used even if Bluetooth is enabled or not

### DIFF
--- a/app/src/main/java/com/nilhcem/blenamebadge/ui/message/MessageActivity.kt
+++ b/app/src/main/java/com/nilhcem/blenamebadge/ui/message/MessageActivity.kt
@@ -237,7 +237,7 @@ class MessageActivity : AppCompatActivity() {
             prepareForScan()
         }
         builder.setNegativeButton("CANCEL") { _, _ ->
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/nilhcem/blenamebadge/ui/message/MessageActivity.kt
```diff
@@ -236,8 +236,8 @@ class MessageActivity : AppCompatActivity() {
         builder.setPositiveButton("OK") { _, _ ->
             prepareForScan()
         }
-        builder.setNegativeButton("EXIT") { _, _ ->
-            finish()
+        builder.setNegativeButton("CANCEL") { _, _ ->
+            Toast.makeText(this, "Enable bluetooth to use send feature", Toast.LENGTH_SHORT).show()
```
yashk2000: Do not use hard-coded strings. Use `enable_bluetooth" from string resources.

Issue #152, opened 2019-04-04
Adding the remaining animations


adityastic:
**Feature Description**
Current Animations are not working in preview:
- Animation
- Laser
- Snowflake
- Picture

**Animations I plan to work on:**
- [x] Animation
- [x] Laser

**Animations @anhanh11001 will to work on:**
- [x] Snowflake
- [x] Picture

**Would you like to work on the issue?**
Yes

anhanh11001:
Are you still working on this or can we split the work in here so that I can work on making 1 or 2 animations?

adityastic:
@anhanh11001 Currently 4 animations are remaining:
Animation
Laser
Picture
Snowflake

Let's divide it. I can work on Animation and Laser and you can work on Picture and Snowflake

adityastic:
@anhanh11001 are you planning to work on Picture animation mode?

anhanh11001:
@adityastic will send the PR tomorrow, I was in the test week last week so I didn't have time to do anything, will continue to work from now on!!

adityastic:
@anhanh11001 great

PR #177, opened 2019-04-09
feat: Animation Mode in Preview

adityastic:
Partially Fixes: #152 (Animation is Checked in the Task List)

Changes: 
- countFrame, lastFrame now further divide the animation into two more steps.
- So we look for not similar frames and update the lastFrame accordingly.
- The Animation is now divided into further 3 parts: 
    * First Phase `countFrame < (badgeWidth / 2)`
    * Second Phase `countFrame < (badgeWidth / 2) & countFrame >  3*(badgeWidth / 2)`
    * Third Phase `countFrame > 3*(badgeWidth / 2)`
- Next we check the lines and the rows lying in that range and we get the desired animation

Screenshots for the change:
animode

### DIFF
--- a/app/src/main/java/com/nilhcem/blenamebadge/ui/badge_preview/PreviewBadge.kt
+++ b/app/src/main/java/com/nilhcem/blenamebadge/ui/badge_preview/PreviewBadge.kt
@@ -51,6 +51,9 @@ class PreviewBadge : View {
     private var checkList = ArrayList<CheckList>()
     private var valueAnimator: ValueAnimator? = null
 
+    private var countFrame = 0
+    private var lastFrame = 0
+
     private fun resetCheckList() {
         checkList = ArrayList()
         for (i in 0 until badgeHeight) {
@@ -111,6 +114,14 @@ class PreviewBadge : View {
             this.badgeSpeed = currentState.getInt(BUNDLE_SPEED)
             this.badgeMode = Mode.values()[currentState.getInt(BUNDLE_MODE)]
             this.checkList = currentState.getParcelableArrayList(BUNDLE_CHECKLIST)
+
+            valueAnimator?.cancel()
+
+            countFrame = 0
+            lastFrame = 0
+
+            configValueAnimation(checkList[0].list.size * 200)
+
             currentState = currentState.getParcelable(BUNDLE_STATE) as Parcelable
         }
         super.onRestoreInstanceState(currentState)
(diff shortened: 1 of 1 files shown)


adityastic:
@yashk2000 @yatri1609 @anhanh11001 Please Review

adityastic:
@anhanh11001 I like the way you optimized the code, please take a look if you can cotribute

anhanh11001:
For naming, I normally avoid making variables' names that contain number like line1, line2. I would do something like firstLine, secondLine.

There is a part of condition looking like this:
```
when {
countFrame < (badgeWidth / 2) || countFrame > (3 * (badgeWidth / 2)) -> j == line1 || j == line2
else -> false
}
```
you can make it a boolean variable with a name like `checkSomething`, it would make the whole condition shorter and more readable in my opinion

adityastic:
@anhanh11001 Great, Let me make those changes

adityastic:
@iamareebjamal Please review

adityastic:
@iamareebjamal Please Have a look

app/src/main/java/com/nilhcem/blenamebadge/ui/badge_preview/PreviewBadge.kt
```diff
@@ -110,7 +114,17 @@ class PreviewBadge : View {
...
             this.badgeSpeed = currentState.getInt(BUNDLE_SPEED)
             this.badgeMode = Mode.values()[currentState.getInt(BUNDLE_MODE)]
+            this.countFrame = currentState.getInt(BUNDLE_CNTFRAME)
+            this.lastFrame = currentState.getInt(BUNDLE_LSTFRAME)
             this.checkList = currentState.getParcelableArrayList(BUNDLE_CHECKLIST)
+
+            valueAnimator?.cancel()
+
+            countFrame = 0
+            lastFrame = 0
```
iamareebjamal: Why?

app/src/main/java/com/nilhcem/blenamebadge/ui/badge_preview/PreviewBadge.kt
```diff
@@ -141,6 +155,8 @@ class PreviewBadge : View {
         bgBounds = RectF((left + offset).toFloat(), (top + offset).toFloat(), (right - offset).toFloat(), ((singleCell * badgeHeight) + (offsetXToAdd * 3)).toFloat())
     }
 
+    private var countFrame = 0
+    private var lastFrame = 0
```
iamareebjamal: Move declaration to top

app/src/main/java/com/nilhcem/blenamebadge/ui/badge_preview/PreviewBadge.kt
adityastic: I am resetting the valueanimator, because both the animators were controlling the animation

adityastic: CountFrame and Lastframe are used to calculate the change in frame, as we are calculating the frames from the divisor and as the number is large, the divisor is same most of the time, these two variables make us check only for the next new number

app/src/main/java/com/nilhcem/blenamebadge/ui/badge_preview/PreviewBadge.kt
adityastic: On it

app/src/main/java/com/nilhcem/blenamebadge/ui/badge_preview/PreviewBadge.kt
adityastic: So I am keeping track of the last count and the frame to start the animation from the exact frame

iamareebjamal: You took them from the bundle and then reassigned to 0

adityastic: Omg, I'm so sorry

adityastic:
@iamareebjamal Fixed

app/src/main/java/com/nilhcem/blenamebadge/ui/badge_preview/PreviewBadge.kt
adityastic: Let me not store them anymore and restart the animation, that would be better

PR #204, opened 2019-04-22
feat: save, export and import badge options 

### LINKED ISSUE #15: Implement an in-app saving and export function for badges
Currently badges cannot be saved. Please add a save button. It would make it easier if the user does not necessarily has to add a name. However the name of the saved file  - it should be possible to change it. Also ensure the user can export the saved file and open it e.g. on another phone.

adityastic:
Fixes #15 
Changes: 
- Save Shapes or Texts
- Sharing Shapes or Texts with Friends using any medium that supports txt files
- Import Shared Files using the + button in the action bar
- A Separate Section to Handle Saved or Imported files
- Send Button is now a Floating Button which works according to the selected viewpager
- Chips denote the attributes of the saved config
- Now you can import files directly from any file browser or any messaging app. Added mimetype to text/plain.
- Checking Overridden Files
- Asking user consent before performing import operations

Screenshots for the change:


adityastic:
@yashk2000 Please Review

adityastic:
@anhanh11001 Please Review

app/src/main/java/com/nilhcem/blenamebadge/util/StorageUtils.kt
```diff
@@ -0,0 +1,99 @@
...
+        saveFile.writeText(json)
+    }
+
+    fun getAllFiles(): List<ConfigInfo> {
+        checkDirectory()
+        val list = mutableListOf<ConfigInfo>()
+
+        val files = File(EXTERNAL_STORAGE_DIRECTORY).listFiles()
+        for (i in files.indices) {
+            if (getFileExtension(files[i].name) == ".txt") {
```
yashk2000: You can use `BADGE_EXTENSION` here.

app/src/main/java/com/nilhcem/blenamebadge/util/StorageUtils.kt
```diff
@@ -0,0 +1,99 @@
...
+        return "$EXTERNAL_STORAGE_DIRECTORY/$fileName"
+    }
+
+    fun copyFileToDirectory(context: Context, uri: Uri?) {
+        checkDirectory()
+        val inputStream = context.contentResolver.openInputStream(uri ?: Uri.EMPTY)
+        val fileName = getFileName(context, uri ?: Uri.EMPTY)
+
+        Timber.e {
+            "fileName $fileName"
```
yashk2000: Is this really needed?

app/src/main/java/com/nilhcem/blenamebadge/util/StorageUtils.kt
adityastic: My bad, on it

app/src/main/java/com/nilhcem/blenamebadge/util/StorageUtils.kt
adityastic: Gotcha

adityastic:
@yashk2000 have a look. Fixed them

adityastic:
@iamareebjamal Please Review sir

adityastic:
@iamareebjamal have a look sir

adityastic:
@iamareebjamal peer review done sir, please have a look

iamareebjamal:
As it is a big PR, it'll take some time to review

adityastic:
@iamareebjamal no worries sir, I can wait. I will work on something else till then

adityastic:
@iamareebjamal BTW sir, I just pushed another commit which performs checks for the saved files to make sure they are not overridden and asks for users choice on the same, As earlier I was saving the file regardless of the user choice. Also fixed a bug which occurs when you try to import any other txt file which does not contain the badge configuration json

adityastic:
Another Commit, Final Changelog:
- Now you can import files directly from any file browser or any messaging app. Added mimetype to text/plain.
- Checking Overridden Files
- Asking user consent before performing import operations  

 @yashk2000 Please test these commits

mariobehling:
Good job. Let's see how this works in everyday usage.

PR #182, opened 2019-04-11
Make preview screen scrollable in landscape mode

yatri1609:
Fixes #168 

Changes: 
Made preview of the badge fixed in portrait mode for the ease of the user. Although it works as it was in the landscape mode. 

Screenshots for the change:
ezgif com-video-to-gif (1)

PR #206, opened 2019-04-24
fix: Deleted unused file which was causing the build to fail and added landscape layout for main_activity

### LINKED ISSUE #205: Build of the app is failing.
**Describe the bug**
The build of the app is failing.

YES.

yashk2000:
Fixes #205 

Changes: Deleted the unused xml file, `message_activity.xml(land)` which was causing build failure.


yashk2000:
@adityastic please review.

anhanh11001:
To handle this, you actually don't need to delete the whole file. The problem is the missing string resource in line 188 of that XML file. So it can be fixed just by adding the string. This is due to #182 

But I'm not sure what's the use of this file in the first place if deleting it doesn't affect the app right?

yashk2000:
The file is now useless. I will be updating this pr. We need to add a landscape layout for the main activity and delete this redundant file. I will update the pr with the changes soon.

adityastic:
@anhanh11001 my last PR Changes the App layout completely, there is no message activity in the first place. The layout needs to be updated with the new layout. So,@yashk2000 I will suggest you to update the landscape layout with the new app layout

yashk2000:

app/src/main/res/layout-land/activity_main.xml
```diff
@@ -0,0 +1,45 @@
...
+            android:layout_height="wrap_content"
+            android:orientation="vertical">
+
+            <com.nilhcem.blenamebadge.ui.badge_preview.PreviewBadge
+                android:id="@+id/preview_badge"
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content" />
+
+            <com.nilhcem.blenamebadge.ui.view.CustomViewPager
+                android:id="@+id/viewPager"
```
anhanh11001: I think you should be careful about naming the id like this. Normally, it should be fine but let's say if you use PreviewBadge View in more than one layout, and you forgot to make a different id name then weird behavior could happen because Android doesn't refer to the right view. Actually, I also saw some bad id naming in the codebase like +@id/text for TextView or +id/recycler_view for RecyclerView

adityastic: That is me, my bad

adityastic:
@yashk2000 please change the title of the pr

yashk2000:
@iamareebjamal please review.

Issue #211, opened 2019-04-25
App crash on saving in horizontal screen


anhanh11001:
**Describe the bug**
On horizontal screen, app crash when users try to save a configuration

**Screenshots**
<img src="https://i.ibb.co/VLZQjWt/ezgif-4-ab2d37d00807.gif" alt="ezgif-4-ab2d37d00807" border="0">

**Would you like to work on the issue?**
Yes

adityastic:
@anhanh11001 any updates on the picture animation?

anhanh11001:
Actually I'm having a bit of problem on that, it is pretty hard for me, I'm doing some refractoring with the Badge Preview class

anhanh11001:
I'm in class but I will private message you later to discuss on that

adityastic:
Cool

PR #212, opened 2019-04-25
fix: Avoid app crash on saving after screen rotation

anhanh11001:
Details:
- App crashes because when the screen is rotated, the recycler view is destroyed. This is avoided by setting retainInstance to true in the onCreate() of the fragment
- Changing same id name for different recyclerView
- Currently, saving a configuration is in onPostExecute() of the AsyncTask class, which means that action saving is done in the main thread. I put it back into the doInBackground()

Fixes: #211

### DIFF
--- a/app/src/main/java/com/nilhcem/blenamebadge/ui/fragments/MainTextDrawableFragment.kt
+++ b/app/src/main/java/com/nilhcem/blenamebadge/ui/fragments/MainTextDrawableFragment.kt
@@ -33,7 +33,6 @@ import com.nilhcem.blenamebadge.util.Converters
 import com.nilhcem.blenamebadge.util.MoshiUtils
 import com.nilhcem.blenamebadge.util.StorageUtils
 import kotlinx.android.synthetic.main.fragment_main_text.*
-import java.lang.ref.WeakReference
 import java.text.SimpleDateFormat
 import java.util.Locale
 import java.util.Calendar
@@ -73,7 +72,7 @@ class MainTextDrawableFragment : BaseFragment() {
         setPreview()
     }
 
-    fun selectText() {
+    private fun selectText() {
         val (valid, textToSend) = Converters.convertTextToLEDHex(if (text_to_send.text.isNotEmpty()) text_to_send.text.toString() else if (!invertLED.isChecked) " " else "", invertLED.isChecked)
         if (!valid) {
             Toast.makeText(context, R.string.character_not_found, Toast.LENGTH_SHORT).show()
@@ -87,7 +86,7 @@ class MainTextDrawableFragment : BaseFragment() {
         )
     }
 
-    fun selectDrawable(selectedItem: DrawableInfo?) {
+    private fun selectDrawable(selectedItem: DrawableInfo?) {
         listener?.onPreviewChange(
             if (selectedItem != null)
                 Converters.convertDrawableToLEDHex(selectedItem.image, invertLED.isChecked)
@@ -193,8 +192,8 @@ class MainTextDrawableFragment : BaseFragment() {
     }
 
     private fun setupRecycler() {
-        recycler_view.layoutManager = LinearLayoutManager(context, LinearLayoutManager.HORIZONTAL, false)
-        recycler_view.adapter = null
+        drawablesRecyclerView.layoutManager = LinearLayoutManager(context, LinearLayoutManager.HORIZONTAL, false)
+        drawablesRecyclerView.adapter = null
 
         val listOfDrawables = listOf(
             DrawableInfo(resources.getDrawable(R.drawable.apple)),
@@ -217,7 +216,7 @@ class MainTextDrawableFragment : BaseFragment() {
         )
 
         drawableRecyclerAdapter = DrawableAdapter(context, listOfDrawables)
-        recycler_view.adapter = drawableRecyclerAdapter
+        drawablesRecyclerView.adapter = drawableRecyclerAdapter
     }
 
     override fun onResume() {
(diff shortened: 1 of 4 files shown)


anhanh11001:
Actually, to fix this issue, I only need to put these 4 lines of code. Others are just changes to improve the codebase.
```
override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
     retainInstance = true
}
```
From what I understand, your suggestion means that if recycler_view is null then we don't update the list at all, which I think is more like avoiding the issue, not really fixing it.

adityastic:
> Actually, to fix this issue, I only need to put these 4 lines of code. Others are just changes to improve the codebase.
> 
> ```
> override fun onCreate(savedInstanceState: Bundle?) {
>      super.onCreate(savedInstanceState)
>      retainInstance = true
> }
> ```
> From what I understand, your suggestion means that if recycler_view is null then we don't update the list at all, which I think is more like avoiding the issue, not really fixing it.

So you're just retaining the fragment which will avoid it from being recreated. Cool.

app/src/main/java/com/nilhcem/blenamebadge/ui/fragments/MainTextDrawableFragment.kt
```diff
@@ -314,35 +314,41 @@ class MainTextDrawableFragment : BaseFragment() {
         return inflater.inflate(R.layout.fragment_main_text, container, false)
     }
 
-    class StoreAsync(private val filename: String, private val json: String, private val context: WeakReference<Context>, private val listener: PreviewChangeListener) : AsyncTask<Void, Void, Boolean>() {
+    inner class StoreAsync(private val filename: String, private val json: String, private val context: WeakReference<Context>, private val listener: PreviewChangeListener) : AsyncTask<Void, Void, Boolean>() {
         override fun doInBackground(vararg params: Void?): Boolean? {
-            return (!StorageUtils.checkIfFilePresent(filename))
+            return when (StorageUtils.checkIfFilePresent(filename)) {
```
adityastic: why did you change it?

app/src/main/java/com/nilhcem/blenamebadge/ui/fragments/MainTextDrawableFragment.kt
```diff
@@ -314,35 +314,41 @@ class MainTextDrawableFragment : BaseFragment() {
...
+                true -> {
+                    saveFile()
+                    true
+                }
+                false -> showFileOverrideDialog()
```
adityastic: very vague, wont work.

app/src/main/java/com/nilhcem/blenamebadge/ui/fragments/MainTextDrawableFragment.kt
```diff
@@ -314,35 +314,41 @@ class MainTextDrawableFragment : BaseFragment() {
...
         }
 
         private fun saveFile() {
             StorageUtils.saveFile(filename, json)
-            listener.updateSavedList()
         }
 
-        private fun showFileOverrideDialog() {
+        private fun showFileOverrideDialog(): Boolean {
+            var saveSuccess = false
```
adityastic: Why are you initialising new variables?

app/src/main/java/com/nilhcem/blenamebadge/ui/fragments/MainTextDrawableFragment.kt
```diff
@@ -314,35 +314,41 @@ class MainTextDrawableFragment : BaseFragment() {
         return inflater.inflate(R.layout.fragment_main_text, container, false)
     }
 
-    class StoreAsync(private val filename: String, private val json: String, private val context: WeakReference<Context>, private val listener: PreviewChangeListener) : AsyncTask<Void, Void, Boolean>() {
+    inner class StoreAsync(private val filename: String, private val json: String, private val context: WeakReference<Context>, private val listener: PreviewChangeListener) : AsyncTask<Void, Void, Boolean>() {
         override fun doInBackground(vararg params: Void?): Boolean? {
-            return (!StorageUtils.checkIfFilePresent(filename))
+            return when (StorageUtils.checkIfFilePresent(filename)) {
+                true -> {
```
adityastic: Please revert StoreAsync to its original state. This is messed up, as far as I know this logic will not work

app/src/main/java/com/nilhcem/blenamebadge/ui/fragments/MainTextDrawableFragment.kt
anhanh11001: From what I understand, you made the StoreAsync class to make the action of saving a new configuration in a different thread. Is it correct?

adityastic: yep correct

anhanh11001: well then currently, the function saveFile() is called in onPostExecute(), which is in the main thread. What I did was to move in into the different thread and then check if action saving is successful then   call listener.updateSavedList()

adityastic: yeah you're completely right on that. But you're calling `showFileOverrideDialog()` in doInBackground. It just wont work, it is not in the main thread and hence will not have the application context with it

adityastic: and also, the logic is the opposite, you should save the file if it did not exist. It is the opposite right now. Have a look there

anhanh11001: oh right that's true, I'll send update later

app/src/main/java/com/nilhcem/blenamebadge/ui/fragments/MainTextDrawableFragment.kt
```diff
@@ -61,71 +60,8 @@ class MainTextDrawableFragment : BaseFragment() {
         }
     }
 
-    override fun getSendData(): DataToSend {
```
adityastic: It is removed and copied at the bottom. Please fix this

anhanh11001: Fixed

PR #218, opened 2019-04-25
feat: Bottombar for navigation between fragments

### LINKED ISSUE #208: Navigation between fragments
Adding Bottombar for navigation between text/drawable and saved fragments

**Screenshots**
Here is a demo with bottombar.
bottombadge

**Additional context**
Currently the user will not recognise if there is another section which can be accessed by swiping towards the left. Therefore, we need to add some kind of navigation option

YES

@iamareebjamal after adding the save feature the users might get confused if there is another section for saved list. The idea of having a bottom bar as it is convenient and has benefits of being right under the thumb. Let me know your thoughts on it sir, I have the PR ready, will proceed if you agree

adityastic:
Fixes #208 

Changes:
- Added a BottomBar for navigation between fragments

Screenshots for the change:


adityastic:
@yashk2000 @anhanh11001 Please Review

adityastic:
@iamareebjamal Please Review

PR #223, opened 2019-04-27
fix: viewpager not initializing when bluetooth not enabled

### LINKED ISSUE #220: Viewpager is not visible when bluetooth permission is denied
**Describe the bug**
On Denying the permission to Enable Bluetooth, we should still be able to use the application. Whereas, Sending should be denied when permission not given  

**To Reproduce**
1. Open the Application without turning the bluetooth ON
2. deny the bluetooth permission

**Expected behavior**
It should be usable by the user to share and preview configurations. 

YES

adityastic:
Fixes #220

Changes: 
- separated `ensureBluetoothEnabled()` from `prepareScan()` 
- fab listener is set according to the bluetooth state

Screenshots for the change:
issue220


adityastic:
@yashk2000 @anhanh11001 Please Review

PR #89, opened 2019-03-26
Code to Convert any 40x11 Translucent Gray-Scaled Bitmap to Hex Arrays

adityastic:
Fixes #17 
Changes: 
- a singleton to convert drawable to hex arrays

Screenshots for the change:
Screenshot_20190325-211831_Badge Magic
Screenshot_20190326-105313_Badge Magic

PR #226, opened 2019-04-27
feat: Added more symbols.

### LINKED ISSUE #98: Implement more symbols.
Some symbols such as cent, yen, pound are not yet implemented in the app. I would like to implement these symbols.

**Additional context**
I would also implement any other missing symbols that I find.

YES.

yashk2000:
Fixes #98 

Changes: More symbols have been implemented. Now, most of the common symbols in the keyboard are in the app.
If any more symbols are found, which are not implemented yet, I shall add them later.


yashk2000:

anhanh11001:
Can I ask how do you find the hex string for the symbol like '¶' equal to "003E7A7A7A3A1A0A0A0A00"? Because if we can find a way to have some kind of function to convert symbol to hex string, it seems better then keep adding hard-coded strings like this

yashk2000:
@anhanh11001 when we click send, the hex format in which the data is being sent is displayed in the activity log. So @adityastic and I created bitmaps of these unimplemented symbols, added them to the recycler view and got the hex form of the symbols in clicking the send. 
There is no direct function in the app to convert a symbol directly to it's hex string.

adityastic:
@anhanh11001 check my PR #89. You need to have the image if you want to convert it to hexadecimal. Think of Text and Drawable as Both Implement the Base as Hexadecimal Strings.

adityastic:
that was the reason, I added the base functionality accordingly specific to this app

yashk2000:
@iamareebjamal please review.

PR #229, opened 2019-04-30
fix: Keyboard goes down when shifting to saved mode.

### LINKED ISSUE #214: Keyboard stays up even on shifting to saved configuration section.
**Describe the bug**
If the keyboard is on the screen while in the text section, and then user shifts to the saved configurations section, the keyboard stays on the screen.

**To Reproduce**
Open the app. The keyboard will already be on the screen. Now directly shift to the saved configuration section. The keyboard does not go down.

**Expected behavior**
The keyboard should go down.

YES.

yashk2000:
Fixes #214 

Changes: Keyboard goes down when shifting to saved mode.


yashk2000:

yashk2000:
@iamareebjamal please review.

PR #257, opened 2019-05-03
fix : Layout Overlap Fix

### LINKED ISSUE #255: Sections are hidden behind save/transfer
Please avoid to implement different layers. The save/transfer buttons should be below the upper area (effects etc). Don't make it go on top. Keep it on the same layer.

Screenshot_2019-05-03-23-33-47-267_com nilhcem blenamebadge

adityastic:
Fixes #255

Changes:
- Moved the layout inside the scrollview

### DIFF
--- a/app/src/main/res/layout/fragment_main_text.xml
+++ b/app/src/main/res/layout/fragment_main_text.xml
@@ -9,138 +9,134 @@
         android:layout_width="match_parent"
         android:layout_height="wrap_content" />
 
-    <RelativeLayout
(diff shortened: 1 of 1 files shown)


yashk2000:
@liveHarshit the UI doesn't look any different. But in devices in which overlapping was happening, it will not happen.

adityastic:
@liveHarshit I don't own a xhdpi device else I could have provided the screenshots. This should fix the issue though

PR #258, opened 2019-05-03
fix : Play Button on Each Tile now plays the badge, transfer is shift…

### LINKED ISSUE #236: Add button "open folder" on top right of screen and delete "+" button (instead add "import" to saved configurations)
1. Please add button "open folder" on top right of screen. This button should direct people to "Saved Configurations"

2. Delete "+" button on the top right of the screen and instead add a button "import" to saved configurations.

3. Add a play button on each tile on the "Saved Config" screen to enable the user to see the saved badges.

4. Change "Saved Configurations" to "Saved Badges" in all views.

5. Delete bottom menu. There is no need for it.

adityastic:
…ed to options now

Fixes #236

Changes: 
- Added Play Pause Button to Saved Badge Tiles
- Changed + Button to Import
- Fixed Issue of Imported Badge not shown

Screenshots for the change:


adityastic:
@yashk2000 @mariobehling @iamareebjamal Please Review Sir

app/src/main/java/com/nilhcem/blenamebadge/ui/drawer/DrawerActivity.kt
```diff
@@ -217,7 +231,7 @@ class DrawerActivity : AppCompatActivity(), DrawerNavigator, NavigationView.OnNa
     }
 
     override fun saveImportFile(uri: Uri?) {
-        if (StorageUtils.copyFileToDirectory(this, uri)) Toast.makeText(this, R.string.success_import_json, Toast.LENGTH_SHORT).show()
```
yashk2000: Is there any reason for removing the toast of successful import? Can't we keep both the toast and the updating of the list?

adityastic: We can, but if the item is successfully imported, it directly shows up in the list. Are you sure the user needs confirmation even when he can see that the file is shown in the list?

yashk2000: I think we can show a toast. If the user has a lot of saved badges, the imported badge might not be immediately visible.

adityastic: @yashk2000 ok, just a min. let me bring it back

adityastic: @yashk2000 Check , Updated

mariobehling:
Thanks. Please follow up on https://github.com/fossasia/badge-magic-android/issues/240 if you like.

adityastic:
@mariobehling definitely sir, it should be done by today

PR #260, opened 2019-05-04
feat : Implement Clipart (Drawables) dropdown area

### LINKED ISSUE #240: Implement Clipart (Drawables) dropdown area
The Cliparts are currently movable to the right and left. Instead there should be an area below and in future the user should be able to add their own cliparts.

Some inspiration from Whatsapp:
Screenshot from 2019-05-01 17-51-20

adityastic:
Fixes #240

Changes:
- Clipart section is now a list with 6 columns

Screenshots for the change:
iss_240

### DIFF
--- a/app/src/main/java/com/nilhcem/blenamebadge/adapter/DrawableItemHolder.kt
+++ b/app/src/main/java/com/nilhcem/blenamebadge/adapter/DrawableItemHolder.kt
@@ -1,5 +1,8 @@
 package com.nilhcem.blenamebadge.adapter
 
+import android.graphics.Bitmap
+import android.graphics.Color
+import android.graphics.drawable.BitmapDrawable
 import android.view.View
 import android.widget.ImageView
 import android.widget.LinearLayout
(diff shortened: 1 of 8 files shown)


adityastic:
@yashk2000 @mariobehling @iamareebjamal Please Review

mariobehling:
Please increase the size of the clipart whereever possible. At the moment some are difficult to see as they are a bit small.

adityastic:
@mariobehling give me a minute

adityastic:
@mariobehling updated the screenshot as well as changed the columns to 5

mariobehling:
Please increase the size of the icons itself inside the white boxes, e.g. make the heart itself bigger.

adityastic:
> Please increase the size of the icons itself inside the white boxes, e.g. make the heart itself bigger.

Sorry for misunderstanding, I got your point. Let me write the algorithm that will scale the images accordingly

adityastic:
@mariobehling Please Look at the new Updated Screenshot Sir. The Algorithm is designed to remove the transparent pixels and scale them to fitCenter inside the White Box

app/src/main/java/com/nilhcem/blenamebadge/adapter/DrawableAdapter.kt
```diff
@@ -50,12 +53,70 @@ class DrawableAdapter(private val context: Context?, private val list: List<Draw
...
                 selectedPosition != -1 && selectedPosition == adapterPosition -> context?.resources?.getDrawable(R.color.colorAccent)
                 else -> context?.resources?.getDrawable(android.R.color.transparent)
             }
+            image.setColorFilter((when {
+                selectedPosition != -1 && selectedPosition == adapterPosition -> context?.resources?.getColor(android.R.color.white)
+                else -> context?.resources?.getColor(android.R.color.black)
+            }) ?: Color.parseColor("#000000"))
+        }
+
+        fun trim(source: Bitmap): Bitmap {
```
iamareebjamal: Should be no need of any algorithm, just reduce the padding on the button. If it is 0, apply negative margin

adityastic: @iamareebjamal ok sir, let me try this as soon as I reach home

adityastic: @iamareebjamal, Btw sir there is no padding given to the button. The drawables are in 40x11. The larger width was the issue which caused the image to fit inside the imageview with the transparent pixels. These transparent pixels made the image look smaller

adityastic: @iamareebjamal I will still have to calculate the amount of margin to reduce for every bitmap as the transparent pixels vary from image to image

adityastic: @iamareebjamal sir, If you look in the current screenshot. Check the clipart at (2,1). It is much wider than the rest. All of them were the same with transparent pixels on all sides, which looked uneven. So I just trimmed those transparent pixels and then set the bitmap.

mariobehling:
The resolution of some images is too small and the icons appear unsharp therefore. You probably need to recreate the images. Just scaling them to another resolution will not solve the problem.

adityastic:
@mariobehling yes sir. The images are only 11 pixels tall. And that is the resolution supported by our badge

adityastic:
@mariobehling I'm traveling right now, will reach by tomm morning. I will resize them to a large scale which will look appealing

adityastic:
@mariobehling Good Morning Sir, I've updated the algorithm thanks for the suggestion they look beautiful now. Please Review :)

app/src/main/java/com/nilhcem/blenamebadge/adapter/DrawableAdapter.kt
adityastic: @iamareebjamal good morning sir, as the padding was 0, I tried applying negative margin of -20dp, It did zoom in but the images which were wider had its end chopped off. So if we plan to give negative margin, it needs to be dynamic according to the transparent pixels for which we need to create an algorithm anyways. After giving the negative margin, the resulting image will be too smooth and hence, we still need to perform a scale which I just added to this function. Please let me know your thoughts on this.
Regards

mariobehling:
Please resolve conflicts.

adityastic:
@mariobehling just a min

app/src/main/java/com/nilhcem/blenamebadge/adapter/DrawableAdapter.kt
adityastic: @iamareebjamal please review

adityastic:
@mariobehling Please Review

mariobehling:
Please follow up here and add more smiley, clock times, thumbs etc. https://github.com/fossasia/badge-magic-android/issues/268

adityastic:
@mariobehling this is a fun one, I'd love to design more cliparts. Thanks for the opportunity

PR #263, opened 2019-05-04
enhancement: Replaced all upper case names

### LINKED ISSUE #250: Change UI to use Capital Letters only for first letter (avoid to capitalize all) and Increase space above Save/Transfer button
1. Please change the UI to use Capital Letters only for first letter (avoid to capitalize all).

2. Increase space above Save/Transfer button in all views. It is currently very close to the above area.

yashk2000:
Fixes #250 

Changes: All existing Upper case strings have been made lower case with only the first letter capital. 

Screenshots for the change:

Screenshot_20190503-232905
Screenshot_20190503-232910


yashk2000:
@mariobehling @iamareebjamal please review.

PR #277, opened 2019-05-11
enhancement: Enhanced UI of the empty saved section.

yashk2000:
Changes: When the saved section is empty, i.e. no badges are saved, a message along with an emoji is displayed.

Screenshots for the change:

Screenshot_20190511-142206


yashk2000:
@adityastic please review. And thanks for providing the emoji :)

app/src/main/java/com/nilhcem/blenamebadge/ui/fragments/main_saved/MainSavedFragment.kt
```diff
@@ -59,6 +59,20 @@ class MainSavedFragment : BaseFragment() {
         return rootView
     }
 
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+
+        if (viewModel.getFiles().value.isNullOrEmpty()) {
+            saved_text.visibility = View.GONE
```
adityastic: Use rootview.saved_text

adityastic:
We should have a saved badges title though, currently only the emoji is visible. Include the title as well which stated 'saved badges'

yashk2000:
> We should have a saved badges title though, currently only the emoji is visible. Include the title as well which stated 'saved badges'

I thought of making it invisible too when there are no badges. I thought it would look better. Tell me if it's fine. If it's not I'll not make it invisible.

app/src/main/java/com/nilhcem/blenamebadge/ui/fragments/main_saved/MainSavedFragment.kt
yashk2000: on it.

adityastic:
@iamareebjamal @mariobehling @CloudyPadmal suggestions plz

mariobehling:
Thank you. Good step forward.

Currently saved badges cannot be edited again. Please help to implement this: https://github.com/fossasia/badge-magic-android/issues/261

PR #304, opened 2019-05-16
fix: memory leak in fragments

### LINKED ISSUE #300: memory leak in maintextfragment
**Describe the bug**
there is a memory leak in maintextfragment due to anonynous listeners

YES

adityastic:
Fixes #300 

Changes: 
- declared all the listeners inside maintextfragment and stopped them accordingly
- removed rootview instance as it caused a major leak in every fragment

### DIFF
--- a/app/src/main/java/org/fossasia/badgemagic/ui/fragments/main_saved/MainSavedFragment.kt
+++ b/app/src/main/java/org/fossasia/badgemagic/ui/fragments/main_saved/MainSavedFragment.kt
@@ -12,6 +12,7 @@ import androidx.core.content.FileProvider
 import androidx.lifecycle.Observer
 import androidx.lifecycle.ViewModelProviders
 import androidx.recyclerview.widget.LinearLayoutManager
+import kotlinx.android.synthetic.main.fragment_main_save.*
 
 import org.fossasia.badgemagic.R
 import org.fossasia.badgemagic.data.fragments.ConfigInfo
@@ -23,7 +24,6 @@ import org.fossasia.badgemagic.ui.fragments.base.BaseFragment
 import org.fossasia.badgemagic.util.Converters
 import org.fossasia.badgemagic.util.InjectorUtils
 import org.fossasia.badgemagic.util.SendingUtils
-import kotlinx.android.synthetic.main.fragment_main_save.view.*
 import org.fossasia.badgemagic.adapter.OnSavedItemSelected
 import org.fossasia.badgemagic.adapter.SaveAdapter
 import java.io.File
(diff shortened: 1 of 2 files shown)


adityastic:
@iamareebjamal @yashk2000 @mariobehling review

iamareebjamal:
And how does this fix the memory leak

adityastic:
@iamareebjamal the textwatcher was getting added everytime when we rotated the screen

adityastic:
@iamareebjamal the textwatcher was getting added everytime when we rotated the screen. As it was an anonymous declaration, it leaked. But now I fixed the scope to the fragment

iamareebjamal:
It is still anonymous

adityastic:
@iamareebjamal How can it be anonymous when I initialized it to a global variable

adityastic:
@iamareebjamal how is it still anonymous when I initialized it to a global variable?. I need to remove the listener as well on 'onStop' dont merge this yet

iamareebjamal:
> I need to remove the listener as well on 'onStop' dont merge this yet

This will stop the memory leak and not making it global. It is still anonymous even if you make it global. Anything implementing SAM without a concrete class is called anonymous inner class. You have just made it static

iamareebjamal:
Sorry, it's not even static. There's literally no change in terms of memory leak and references from what it was before

adityastic:
@iamareebjamal yes I forgot to include the call to the removelistener in onstop. I've added it locally yesterday, will push it once I reach home

adityastic:
@iamareebjamal I get your point with anonymous class declaration. But if I don't initialize it globally I won't be able to remove the listener as removelistener accepts the object which needs to be removed

adityastic:
@iamareebjamal, Damn you are right. The cause was something completely different, thanks for your suggestion man

adityastic:
@iamareebjamal review

iamareebjamal:
Why so many changes?

adityastic:
@iamareebjamal the rootview was leaking

iamareebjamal:
So?

adityastic:
@iamareebjamal so I fixed it by not storing the view itself

iamareebjamal:
Summarize your changes please

adityastic:
- removed rootview instance
- removed listeners onStop and add listeners back onStart

iamareebjamal:
Why do you need to change so many things for this?

iamareebjamal:
Also, not storing a variable does not solve a memory leak. Just removing the listeners did the job

adityastic:
@iamareebjamal there aren't so many changes. Every view was called like this rootView.textview. I just removed rootview from everything

adityastic:
@iamareebjamal after removing the listeners, I got a leak regarding rootview and viewmodels.

iamareebjamal:
Why did you change from lateinit var to nullable? There definitely are many changes unlrelated to memory leak which also reduce the code quality

adityastic:
@iamareebjamal better now?

PR #311, opened 2019-05-21
fix: memory leak in preview badge

adityastic:
Fixes #311

Changes: 
- Listener removed ondetach


adityastic:
@yashk2000 @iamareebjamal @mariobehling Please Review

PR #319, opened 2019-05-22
fix: memory leak due to EditableInputConnection

### LINKED ISSUE #316: memory leak due to EditableInputConnection
**Describe the bug**
memory leak due to EditableInputConnection

**Screenshots**
nmmem

YES

adityastic:
Fixes #316

Changes: 
- fixed memory leak by disabling the InputConnection

Additional Context:
https://stackoverflow.com/questions/14069501/edittext-causing-memory-leak

### DIFF
--- a/app/src/main/java/org/fossasia/badgemagic/ui/custom/NoInputConnEditText.kt
+++ b/app/src/main/java/org/fossasia/badgemagic/ui/custom/NoInputConnEditText.kt
@@ -1,0 +1,41 @@
+package org.fossasia.badgemagic.ui.custom
+
+import android.content.Context
+import android.text.TextUtils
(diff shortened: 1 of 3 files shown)


adityastic:
@iamareebjamal please check the image I posted in the issue. There is an issue with the InputConnection which is responsible for giving suggestions on the keyboard. Those are getting leaked from the edittext, I tried removing the inputText attribute and it still leaked. I am sure I am not working with the InputConnection of the Editext anywhere in the code. Please have a look if you can find the leak

iamareebjamal:
I will. But priority wise. Having a memory leak which is ont our fault is much much better than having to maintain a custom view

PR #332, opened 2019-05-26
chore: updated conditions in automatic release scripts

adityastic:
Changes: 
- Changes to condition


adityastic:
@iamareebjamal have a look

PR #334, opened 2019-05-27
chore: Updated Conditions and bumped up minSDK

adityastic:
Changes:
- Updated Conditions and Bumped up gradle version


adityastic:
@iamareebjamal have a look

PR #359, opened 2019-06-08
chore: Replaced context with requireContext

yashk2000:
Changes: Replaced `context` with `requireContext()` as it is better for use in fragments. `context` might return a null value, but `requireContext()` will never return a null value.


yashk2000:
@adityastic @mariobehling @iamareebjamal @cweitat please review.

PR #418, opened 2019-06-28
fix: force stop when rejected permission

### LINKED ISSUE #416: Badge Magic v1.2.0 - Application suddenly stops when saving "Badge"
# Actual behaviour
It should, though I don't allow applications to access location and storage on the device when starting the application.  This behavior will not stop the application when I save "Badge".

<code>According to my analysis, while storing "badge", the application must request back permission to access the location and storage on the device.</code>

# Expected behaviour
This application suddenly stopped when I saved "Badge". Because before I did not allow applications to access location and storage on the device.

# Step to reproduce
- Install and open the application

- Select "Deny" when the application requests permission to access location and storage.

- Enter teks, Swipe right, select animation "Right". Then click "Save". 

- Then you will find problems like I have experienced.

# Information
- Device: Nubia n2 lite 
- Android version: 8.1.0
- Application version: 1.2.0

# Recording of the bug
https://youtu.be/BPnKGwYzZxI

adityastic:
Fixes #416

Changes: 
- Added Permission Check Before saving badge

Screenshots for the change:

### DIFF
--- a/app/src/main/java/org/fossasia/badgemagic/ui/fragments/TextArtFragment.kt
+++ b/app/src/main/java/org/fossasia/badgemagic/ui/fragments/TextArtFragment.kt
@@ -6,6 +6,7 @@ import android.bluetooth.BluetoothAdapter
 import android.bluetooth.BluetoothManager
 import android.content.Context
 import android.content.DialogInterface
+import android.content.pm.PackageManager
 import android.content.res.Configuration
 import android.graphics.Color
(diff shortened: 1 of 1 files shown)


iamareebjamal:
If user allows the permission, resume from there

adityastic:
> If user allows the permission, resume from there

Sure

adityastic:
@iamareebjamal review

PR #488, opened 2019-09-03
chore: check images for fastlane and docs

### LINKED ISSUE #486: Readme.md: Add screenshots to /docs/images and add them into the Readme
The Readme.md currently contains screenshots from the Fastlane deployment. All documentation for the project itself should be using the /docs folder. Please **add the screenshots** to /docs/images and link those screenshots instead.

adityastic:
Fixes #486 

Changes: 
- Moved Images to docs folder and we now copy it when we publish using fastlane


scripts/upload-apk.sh
```diff
@@ -36,6 +36,11 @@ git push origin apk --force --quiet > /dev/null
 
 if [[ $TRAVIS_BRANCH =~ ^(master)$ ]]; then
     cd ..
+
+    # Copy Screenshots
+    mkdir -p "fastlane/metadata/android/en-US/images/phoneScreenshots"
+    find docs -type f -name "screen-?.png" -exec cp {} "fastlane/metadata/android/en-US/images/phoneScreenshots/" \;
```
iamareebjamal: Just use cp

iamareebjamal:
Need to find a solution to have F-Droid work as well with this. It won't find the images in this case

scripts/upload-apk.sh
adityastic: @iamareebjamal regex does not work in cp

adityastic:
> Need to find a solution to have F-Droid work as well with this. It won't find the images in this case

Yes true, let me look what else can I do

scripts/upload-apk.sh
adityastic: yeah not using regex, glob can work even in cp. Updated

adityastic:
> Git logically stores each file under its SHA1. What this means is if you have two files with exactly the same content in a repository (or if you rename a file), only one copy is stored.

@iamareebjamal we should not have problem duplicating content

iamareebjamal:
Not the case with binaries. But let me check, maybe I am wrong

iamareebjamal:
Yeah, it'll work. Please go ahead. 

However, please fail the build if there are different count of images in fastlane and docs folder so that when an image is added in one of them, it reminds the developer to add image in the other folder as well

adityastic:
@iamareebjamal yeah that sounds right. Will get it done in some time

adityastic:
@iamareebjamal this is the build failing, I did not opt to check the count rather I check the md5 which will give us the exact content duplication difference.

Failing Build Result: https://travis-ci.com/fossasia/badge-magic-android/builds/126048064
Passing Build Result: https://travis-ci.com/fossasia/badge-magic-android/jobs/231440485

scripts/check-screenshots.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+FILES_DIFF=$(diff <(find docs/images -type f -name "screen-?.png" -exec md5sum {} + | sort -k 2 | sed 's/ .*\// /') <(find fastlane/metadata/android/en-US/images/phoneScreenshots -type f -exec md5sum {} + | sort -k 2 | sed 's/ .*\// /'))
+if [[ $FILES_DIFF ]]; then
+    echo -e "\033[0;31mScreenshots in docs/images and fastlane/metadata/android/en-US/images/phoneScreenshots are not same\033[0m" >&2
+    exit 5;
```
iamareebjamal: Why not exit 1?

adityastic: Yes 1 is for general errors, we can use that as well. I just wanted to make it unique to our application so I chose a random 5

iamareebjamal: I think it is better to use standard errors. CI may be hardcoded to understand only 1 exit code

adityastic: @iamareebjamal yes true, updated with `exit 1`

PR #595, opened 2019-12-31
feat: multiplatform module for Android and iOS

### LINKED ISSUE #575: Kotlin Multiplatform Discussion
Shift the model code to a multiplatform kotlin native

adityastic:
Fixes #575


adityastic:
@iamareebjamal I plan on creating another branch to push the latest SharedFramework which we would download when compiling badge-magic-ios , let me know your suggestions on this

iamareebjamal:
Sounds good but why not do it in one repo

adityastic:
> Sounds good but why not do it in one repo

yes, that is an amazing idea, we can shift iOS directly here and maintain both the apps right here itself

adityastic:
@iamareebjamal we would then have to manage travis to use macOS and manually download android sdk, gradle, jdk and the rest of things inside a android docker. Let me know your thoughts, If it is worth the effort

iamareebjamal:
You can use environments to differentiate between both. Not sure how easy it is in Travis but here is an example of GitHub Actions - https://github.com/korlibs/klock

adityastic:
@iamareebjamal the catch with environments is that they run concurrently, we might want to synchronously run our project on an android image to build our iOS-framework then run it on OS X image to build iOS with the built framework.

adityastic:
@iamareebjamal nvm, I managed to sync it HAHA

adityastic:
@iamareebjamal Changes:

1. Shifted some of the logic to shared module.
2. Removed Moshi and used kotlin serializer so now we can handle same classes in any platform.
3. changed travis to build android then iOS.

Next Steps: Add iOS project tomorrow

adityastic:
@iamareebjamal please review

app/src/main/java/org/fossasia/badgemagic/adapter/SaveAdapter.kt
```diff
@@ -10,10 +10,10 @@ import androidx.appcompat.widget.AppCompatImageView
 import androidx.core.content.ContextCompat
 import androidx.recyclerview.widget.RecyclerView
 import com.google.android.material.chip.Chip
+import data.BadgeConfig
+import data.ConfigInfo
+import helpers.JSONHelper
```
iamareebjamal: Why not prepending org.fossasia.badgemagic and using same package structure in common module as well?

adityastic: Yep, it would be perfect

PR #603, opened 2020-01-02
fix: location issue with sending badges

### LINKED ISSUE #583: F-Droid build doesn't work
**Describe the bug**
The current build on F-Droid doesn't seem to work. I tested it with two different Android devices. When clicking "Transfer" to the badge, it complains about `No GPS Found, Please Enable GPS from settings, despite GPS being allowed and activated.

**To Reproduce**
Install the app from F-Droid and try to transfer some content to the badge.

adityastic:
Fixes #583


fcartegnie:
You should not change indentation unrelated to the fix

adityastic:
@fcartegnie it is handled by spotless. It takes care of the code indentation

PR #606, opened 2020-01-03
[WIP] chore: shift ios project to this project

### LINKED ISSUE #604: shifting iOS app here

adityastic:
Fixes #604


adityastic:
@iamareebjamal travis is one hell of a mess. It does not allow you to use android language on macOS image. 
Timeline:
1. Thought of running gradle on macOS by downgrading the jdk to 8. 
2. It was still building the android project even when I was building a specific module with 
```./gradlew BadgeMagicModule:packForXcode```
3. I finally managed to make it work by not including the android project in gradle.settings
4. So, I combined it all and removed the android project on macOS push.

Let me know your thoughts on this approach. This is a workaround and I'm sure there must be a better way or we can switch to Github Workflow which might be better as a whole.

**Comment by adityastic** [comment]
@iamareebjamal if we can manage to build only the module we can solve this without issues, but for some reason even after specifying the module with the command it builds the android project which fails for some unknown reason https://travis-ci.com/fossasia/badge-magic-android/jobs/272179208.

File: .travis.yml [hunk 1]
```diff
@@ -48,14 +48,48 @@ jobs:
...
+        - sudo rm -rf /Library/Java/*
+        - sudo rm -rf /Library/PreferencePanes/Java*
+        - sudo rm -rf /Library/Internet\ Plug-Ins/Java*
+        - brew tap AdoptOpenJDK/openjdk
+        - brew cask install adoptopenjdk8
+        - java -version
+        # -----------
+        # -----------
+        # Remove Project from gradle.settings
+        - echo "include ':BadgeMagicModule'" > settings.gradle
```
adityastic: @iamareebjamal that's the hack for now

iamareebjamal: Works for now

iamareebjamal:
Merge after changing the PR title

iamareebjamal:
The build is failing because gradle is built on JDK 12 and JDK 13 is installed on the platform. gradle only supports JDK <= 12

So, if you can use that, it'll fix the error

adityastic:
@iamareebjamal Yes, Even I thought the same and tried all jdk's from 8 to 12, all failed because of Kotlin NullPointerException in Databinding. Here are the links of all builds, It should not build the android project as I was specifying the module specifically, but for some reason it was.

jdk8 - https://travis-ci.com/fossasia/badge-magic-android/jobs/272161844
jdk9 - https://travis-ci.com/fossasia/badge-magic-android/jobs/272180978
jdk11 - https://travis-ci.com/fossasia/badge-magic-android/jobs/272181067
jdk12 - https://travis-ci.com/fossasia/badge-magic-android/jobs/272181305

Issue #616, opened 2020-01-04
Application directly closes on back pressed


tanmayrauth:
**Describe the feature you'd like**
Application directly closes on back pressed. If an user clicks the back button unintensionally then the application is terminated . A small dialog box can fix this issue.

**Would you like to work on the issue?**
Sure

adityastic:
We might not need this

tanmayrauth:
But you might have seen this feature in many apps where it is used to avoid unexpected back clicks. Isn't it?

iamareebjamal:
Yes, non standard apps. If you made a mistake, you can go back to the app using switcher in 2 clicks. We'll be making users deliberately click twice who want to exit the app. Why can't people who made a mistake click twice?

tanmayrauth:
No worries. Thanks for quick look.

tanmayrauth:
But one thing I have observed is that our app is missing the flow.
There is no home menu. I consider create badge to be home menu.

So if someone clicks back from eg. Settings fragment ideally they should land to home menu. 

But in our current situation if one clicks back from any fragment the app is closed.

Is this expected behaviour? I believe there should be a home menu.

PR #617, opened 2020-01-04
feat: Confirm unexpected back pressed exit #616 

tanmayrauth:
Fixes #[616]

Changes: [ Added a dialog box check to confirm that the back pressed was not unintentional. Many times it happens that user intentionally click on back button and exit the app and they don't return.
This confirmation will help to retain user.

Screenshots for the change:
ezgif com-video-to-gif (2)


tanmayrauth:
@adityastic why is PR closed? Anything missing?

adityastic:
@tanmayrauth we might not need this feature atm tanmay, this would be annoying if someone just wants to exit the app and a dialog pops up on their face. I hope you understand the point, thanks for your contributions

PR #620, opened 2020-01-05
fix: Semantic changes for delete confirmation message

tanmayrauth:
Fixes #[619]

Changes: Corrected delete messages for both languages, similar to that of saved badge message.


iamareebjamal:
Please mark Fixes issue correctly. @adityastic Please ensure this before merging

adityastic:
@iamareebjamal my bad, will take care of it

Issue #660, opened 2020-02-17
When trying to save a new drawing the application closes


carlosnewmusic:
**Describe the bug**
When trying to save a new drawing the application closes

**To Reproduce**
Display vertically, Draw, click on the save icon, and the application closes unexpectedly

**Expected behavior**
That will save the drawing and continue to function normally

**Logs**
How do I get them?

**Smartphone Info:**

|               |   |
|---------------|---|
|Device         | VDF1400 |
|Android Version| 6.0.1   |

refactor-droidyy:
@carlosnewmusic attach gif for the same.

adityastic:
@carlosnewmusic can you test the latest version and report back about the issue?

carlosnewmusic:
> @carlosnewmusic can you test the latest version and reporting back about the issue?

![Screenshot_20200220-114326](https://user-images.githubusercontent.com/19379457/74926635-7ccae580-53d6-11ea-8fb8-c17f4d063bd6.png)
Where I download the application in the f droid repositories the update does not appear

iamareebjamal:
It takes some time for it to appear on F-Droid. Even more given that there was a recent fix in out config 5 days ago https://gitlab.com/fdroid/fdroiddata/-/commit/224a396caf18f883654eef2b2bf31df9da98b890

You can download the latest APKs from here - https://github.com/fossasia/badge-magic-android/tree/apk

PR #662, opened 2020-02-20
chore: fix apk path in script

### LINKED ISSUE #661: fix apk path in script

adityastic:
Fixes #661

### DIFF
--- a/scripts/upload-apk.sh
+++ b/scripts/upload-apk.sh
@@ -17,11 +17,7 @@ fi
 find ../android/build/outputs -type f \( -name '*.apk' -o -name '*.aab' \) -exec cp -v {} . \;
 
 for file in android*; do
-    if [[ $file =~ ".aab" ]]; then
-        mv $file badge-magic-$TRAVIS_BRANCH-$file
-    else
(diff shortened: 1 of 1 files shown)


scripts/upload-apk.sh
```diff
@@ -20,7 +20,7 @@ for file in android*; do
     if [[ $file =~ ".aab" ]]; then
         mv $file badge-magic-$TRAVIS_BRANCH-$file
     else
-        mv $file badge-magic-$TRAVIS_BRANCH-${file:4}
+        mv $file badge-magic-$TRAVIS_BRANCH-${file:8}
```
iamareebjamal: Hmm I thought you would have realized how brittle this is

adityastic: I just noticed that the names are not correct in the apk branch when you posted a link in #660 xD

iamareebjamal: Yes, it is because this method of renaming is error prone

adityastic: Let me find the first index of - and substring. I'll update it in sometime

iamareebjamal (approved):

PR #745, opened 2021-01-06
fix: Cliparts size mismatch with badge size

### LINKED ISSUE #586: Cliparts size is 40x11 even though badge size is 44x11
First up, thanks for your amazing work on this app. Really love the easy way to change badges!
**Describe the bug**
I noticed that cliparts can only be created in a 40x11 px format (stored png also has this format). But the badge has a 44x11 px width. Any reason for that?

**To Reproduce**
Create a new clipart and fill it with pixels. Now go to "create badge" and insert it. There will be an area of 2px left and right of the clipart which will not have any red pixels.

**Expected behavior**
I would expect the canvas for new cliparts to be 44px wide.

**Screenshots**
Photo of the clipart:
photo5208504173140749201
Photo of the badge with padding.
photo5208648853409082519

**Smartphone Info:**
|Device         | Samsung Galaxy S5 (klte) |
|Android Version| 9 (Lineageos)    |
App version: 1.5.0

My android skills are low, so rather not.

akri16:
Fixes #586

Changes: 

1. Change the dimension to 40 x 11 at all places
2. Refactor and fix associated errors

Screenshots for the change:

WhatsApp Image 2021-01-06 at 18 31 00 (1)

WhatsApp Image 2021-01-06 at 18 31 00

### DIFF
--- a/android/src/main/java/org/fossasia/badgemagic/ui/custom/PreviewBadge.kt
+++ b/android/src/main/java/org/fossasia/badgemagic/ui/custom/PreviewBadge.kt
@@ -12,6 +12,7 @@ import android.graphics.drawable.Drawable
 import android.os.Bundle
 import android.os.Parcelable
 import android.util.AttributeSet
+import android.util.Log
 import android.view.View
 import android.view.animation.LinearInterpolator
 import androidx.annotation.Nullable
@@ -20,6 +21,7 @@ import org.fossasia.badgemagic.data.Mode
 import org.fossasia.badgemagic.data.Speed
 import org.fossasia.badgemagic.data.badge_preview.Cell
 import org.fossasia.badgemagic.data.badge_preview.CheckList
+import org.fossasia.badgemagic.util.Converters
 import org.fossasia.badgemagic.util.Converters.hexToBin
 
 private const val BUNDLE_STATE = "superState"
@@ -30,7 +32,7 @@ private const val BUNDLE_MODE = "badgeMode"
 private const val BUNDLE_CHECKLIST = "checkList"
 
 class PreviewBadge : View {
-
+    private val TAG = "PreviewBadge"
     private var ledDisabled: Drawable
     private var ledEnabled: Drawable
 
@@ -38,8 +40,8 @@ class PreviewBadge : View {
     private var cells = mutableListOf<Cell>()
 
     private var badgeHeight = 11
-    private var badgeWidth = 44
-
+    private var badgeWidth = 40
+    private val maxOffset: Int by lazy { Converters.DpToPx(24, context) }
     private var oneByte = 8
 
     private var animationIndex: Int = 0
@@ -91,18 +93,18 @@ class PreviewBadge : View {
     override fun onMeasure(widthMeasureSpec: Int, heightMeasureSpec: Int) {
         super.onMeasure(widthMeasureSpec, heightMeasureSpec)
 
-        val ratioHeight = 1
-        val ratioWidth = 3
-
         val originalWidth = MeasureSpec.getSize(widthMeasureSpec)
-        val calculatedHeight = originalWidth * ratioHeight / ratioWidth
+        val offset = (measuredWidth / 32).run { if (this > maxOffset) maxOffset else this }
+        val singleCell = (originalWidth - 4 * offset) / badgeWidth
+
+        val calculatedHeight = singleCell * badgeHeight + 4 * offset
 
         setMeasuredDimension(
             MeasureSpec.makeMeasureSpec(originalWidth, MeasureSpec.EXACTLY),
-            MeasureSpec.makeMeasureSpec(calculatedHeight.toInt(), MeasureSpec.EXACTLY))
+            MeasureSpec.makeMeasureSpec(calculatedHeight, MeasureSpec.EXACTLY))
     }
 
-    public override fun onSaveInstanceState(): Parcelable? {
+    public override fun onSaveInstanceState(): Parcelable {
         val bundle = Bundle()
         bundle.putParcelable(BUNDLE_STATE, super.onSaveInstanceState())
         bundle.putBoolean(BUNDLE_FLASH, this.ifFlash)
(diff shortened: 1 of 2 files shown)


adityastic:
I guess we have to do the other way around, it has to be 44*11 everywhere. Because the hardware actually has 44x11 leds physically. If you look at the readme you can count the leds in the screenshots

akri16:
I saw 40 x 11 at most of the places that's why I changed the badge preview to 40 x 11 too. But yeah sure I`ll do this. Give me a day I'll send a new PR.

adityastic:
> I saw 40 x 11 at most of the places that's why I changed the badge preview to 40 x 11 too. But yeah sure I`ll do this. Give me a day I'll send a new PR.

Yeah no worries, take your time

akri16:
Actually, there is a problem that I just noticed. There is a padding of 2 columns before and after each drawable. So when opened in FIXED mode it's not possible to display the whole object because of padding. The solution for this is to remove the left padding for the first object and right padding for the last object.

So the problem here is that even if you completely fill the DrawLayout and save the clipart. The preview in fixed mode will have some padding in it because of which it won`t be displayed completely.

**You can see in this image, the first two columns are empty**

WhatsApp Image 2021-01-07 at 16 00 19

I just wanna know if this is the intended behavior or is it a bug. In case of the latter, please also suggest what is to be done.

adityastic:
> I just wanna know if this is the intended behavior or is it a bug. In case of the latter, please also suggest what is to be done.
> 

So in order to understand this we need to look at how the badge actually works. It takes in a series of hexadecimal strings which represent the pixels that need to be turned on or off. Every hex string is supposed to be of length 22. A nice example can be:

Hex String: 00000000007E0000000000

Rendering in Binary:
0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0111 1110 0000 0000 0000 0000 0000 0000 0000 00000000 0000

Rendering in Badge:
| 0000 0000 |
| 0000 0000 |
| 0000 0000 |
| 0000 0000 |
| 0000 0000 |
| 0111 1110 |
| 0000 0000 |
| 0000 0000 |
| 0000 0000 |
| 0000 0000 |
| 0000 0000 |

I hope you can see the - symbol being created with the 1's. So a minimum length you can send is 22. 

Now if you want to increase the width of this - symbol and make it like this.

| 0000 0000 | 0000 |
| 0000 0000 | 0000 |
| 0000 0000 | 0000 |
| 0000 0000 | 0000 |
| 0000 0000 | 0000 |
| 0111 1111 | 1110 |
| 0000 0000 | 0000 |
| 0000 0000 | 0000 |
| 0000 0000 | 0000 |
| 0000 0000 | 0000 |
| 0000 0000 | 0000 |

Single block...(^)

You cannot send a single block of length 11 or 4 columns. It has to be 22. So in order to increase the length of the - symbol you have to send something with 16 columns like this.

| 0000 0000 | 0000 0000 |
| 0000 0000 | 0000 0000 |
| 0000 0000 | 0000 0000 |
| 0000 0000 | 0000 0000 |
| 0000 0000 | 0000 0000 |
| 0111 1111 | 1110 0000 |
| 0000 0000 | 0000 0000 |
| 0000 0000 | 0000 0000 |
| 0000 0000 | 0000 0000 |
| 0000 0000 | 0000 0000 |
| 0000 0000 | 0000 0000 |

This the extra padding you are seeing there. When you create the ❤️ symbol there it will be using more than 8 columns hence it will automatically adjust into the next multiple of 8 that is 16 ( just like the above example of - symbol) it expands. 

Hope you now see why this extra padding is added and why it is the intended behavior to make cliparts function the way they do right now. They are centered to make it look good by my algorithm. 


**Possible Solutions if you want multiple hearts without padding**
Not use a single heart clipart and create a clipart with three hearts inside it.
Or
Create a heart with only 8 columns.


> So the problem here is that even if you completely fill the DrawLayout and save the clipart. The preview in fixed mode will have some padding in it because of which it won`t be displayed completely.
>

If you take the same concept of applying padding and expand it to multiple hex strings. The number of columns possible in draw layout are 40, which is a multiple of 8. Hence no padding is applied to them. As you plan to increase the number from 40 to 44, you will have to remove the algorithm to add padding on the left and right sides to make the element look like it is centered( can be found in Converters, not sure though), and add all the padding at the end of the hex strings. So in theory you will be making the cliparts of size 48x11 generated from the 44x11 drawlayout( because it the columns have to be a multiple of 8 in order to be read by the actual hardware as example shown above).

Edit: Sorry for the long reply, I just love bitmaps 😂. Hope this gives some light on how the core functions work.

Kamal-Sonani:
https://play.google.com/store/apps/details?id=justsw.tonypeng.ledsigns

Just like this app

As single color data 

Binary String example for 5 x 5 matrix ,only border glow

11111
10001
10001
10001
11111

But for multi color ,looks like this
Red color border (with no anti alias as depands on emoji or entered text )
FF0000 FF0000 FF0000 FF0000 FF0000
FF0000 000000 000000 000000 FF0000
FF0000 000000 000000 000000 FF0000
FF0000 000000 000000 000000 FF0000
FF0000 FF0000 FF0000 FF0000 FF0000

akri16:
@adityastic Thanks for the extensive explanation! Had to go through it a couple of times to get my head around it. But it definitely helped me to get the big picture. 

So I understand that to change this behavior, a lot of changes in the algorithm will have to be done.  Also, this is somewhat related to how I am planning to add the Rich Text Editor.  As the issue that we are trying to solve here didn't create nor alters the padding issue, why don`t we open a fresh issue and solve it with a brand new PR? 

Regarding this issue, I've made the necessary changes and created a new PR. Please review and merge #746

_Also question:_ Is the app configured to send the data as HEX because that`s how the API Contract is (Like, how the hardware expects it)? Or is it just one way of doing it? And is there any limit for the number of HEXs that can be sent to the hardware before it begins to forget the previous data?

PS: Reason for the late reply: Have exams these days :)

adityastic:
>why don`t we open a fresh issue and solve it with a brand new PR? 
> 
Yeah feel free to open issues, I'm always available for a conversation.

> Is the app configured to send the data as HEX because that`s how the API Contract is (Like, how the hardware expects it)? Or is it just one way of doing it? 
>
I dont really know much about the hardware. All the information available is gathered using wireshark. 

>And is there any limit for the number of HEXs that can be sent to the hardware before it begins to forget the previous data?
> 
Not as far as I know of. I haven't stress tested the badge nor have the plans too. If we will have to we will eventually. But no plans as of yet.

> PS: Reason for the late reply: Have exams these days :)
>
Exams should have greater importance so keep open source casual, give higher priority to studies.

Issue #948, opened 2024-06-06
Implement Logic to display clipart to the physical badge.


Jhalakupadhyay:
**Describe the feature you'd like**
The user should be able to send the clipart to the physical badge by selecting the vector.

**Would you like to work on the issue?**
I will be working on it this week.

PR #958, opened 2024-06-28
feat: Added the Logic for changing clipart to led hex

Jhalakupadhyay:
#948 

- Added the logic to display the clipart to the physical badge.
- Added the test cases for the DataToByteArray converter class.


adityastic (approved):
LGTM!

Issue #959, opened 2024-06-29
Create the state Machine for the BLE transfer.


Jhalakupadhyay:
currently, when clicking on Bluetooth transfer we get only two messages transferring.. and successfully transfers but we should be able to get toast according to the state that the machine is in and any issue on that particular state should be handled gracefully.

PR #960, opened 2024-06-30
feat: Added the BLE state machine to handle each transfer state gracefully.

Jhalakupadhyay:
#959 Before the whole logic for transfer of data was in one file now the logic is divided into different states and each state is handled an according to the success and failure of the state toast is shown on the device.

### DIFF
--- a/lib/constants.dart
+++ b/lib/constants.dart
@@ -1,3 +1,5 @@
+const homeScreenTitleKey = "bm_hm_title";
+
 //path to all the animation assets used
 const String animation = 'assets/animations/ic_anim_animation.gif';
 const String aniLeft = 'assets/animations/ic_anim_left.gif';
--- a/.github/actions/screenshot-android/action.yml
+++ b/.github/actions/screenshot-android/action.yml
@@ -1,0 +1,67 @@
+name: "Android Screenshots Workflow"
+
+inputs:
+  ANDROID_EMULATOR_API:
+    description: 'Emulator API to be used when running tests'
+    required: false
+    default: 34
+  ANDROID_EMULATOR_ARCH:
+    description: 'Emulator architecture to be used when running tests'
+    required: false
+    default: x86_64
+
+runs:
+  using: "composite"
+  steps:
+    - name: Enable KVM group perms
+      shell: bash
+      run: |
+        echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+        sudo udevadm control --reload-rules
+        sudo udevadm trigger --name-match=kvm
+
+    - name: Cache AVD
+      uses: actions/cache@v4
+      id: avd-cache
+      with:
+        path: |
+          ~/.android/avd/*
+          ~/.android/adb*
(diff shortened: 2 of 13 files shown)


lib/providers/badge_message_provider.dart
```diff
@@ -34,25 +38,89 @@ class BadgeMessageProvider extends ChangeNotifier {
...
+          ],
+        ),
+        backgroundColor: Colors.white,
+        behavior: SnackBarBehavior.floating,
+        shape: RoundedRectangleBorder(
+          borderRadius: BorderRadius.circular(10),
+        ),
+        dismissDirection: DismissDirection.startToEnd,
+      ),
+    );
```
adityastic: Should be part of Scanning state

lib/bademagic_module/bluetooth/scanstate.dart
```diff
@@ -0,0 +1,62 @@
...
+  @override
+  Future<BleState?> isFailed(String message) async {
+    toast.failureToast(message);
+    return null;
+  }
+
+  @override
+  Future<BleState?> isSuccess(String message) async {
+    toast.successToast(message);
+    return ConnectState(scanResult: foundDevice!).processState();
```
adityastic: You can have a cleaner implementation by sending the state back to the main thread and let it execute until completion instead of having a nested call

lib/bademagic_module/bluetooth/scanstate.dart
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  @override
+  Future<BleState?> processState() async {
```
adityastic: You're not making use of the returned BLEState. Every state returns null. I'd advise cleaning this up and returning the next state rather than making nested calls. Letting the main thread run until completion

adityastic (changes requested):

lib/constants.dart
```diff
@@ -12,3 +12,15 @@ const String aniRight = 'assets/animations/ic_anim_right.gif';
...
+//enum to manage the state of the bluetooth connection
+enum TransferState {
+  search,
+  connecting,
+  connected,
+  notConnected,
+  transfer,
+  permission,
+  success,
+  failed,
```
adityastic: Can be removed

lib/bademagic_module/bluetooth/ble_state_interface.dart
```diff
@@ -0,0 +1,7 @@
+abstract class BleState {
+  Future<BleState?> isSuccess(String message);
+
+  Future<BleState?> isFailed(String message);
+
```
adityastic: These method can be improvised or removed. Showing Success or Failed Toast can be a Completed State. That state shows a toast

lib/bademagic_module/bluetooth/bletoast.dart
```diff
@@ -0,0 +1,73 @@
...
+        shape: RoundedRectangleBorder(
+          borderRadius: BorderRadius.circular(10),
+        ),
+        dismissDirection: DismissDirection.startToEnd,
+      ),
+    );
+  }
+
+  //show toast for failure state of BLE
+  void failureToast(String message) {
```
adityastic: This looks like a duplicate of the above. Why not reuse the function?

PR #967, opened 2024-07-04
fix: overhaul file handling and sharing badges

kleinesfilmroellchen:
*_EXTERNAL_STORAGE is deprecated and non-functional in Android 13+. There really isn't any reason to use external storage for the badges and cliparts anyways. This commit changes Badge Magic to use app-internal internal storage that can be freely accessed without requiring permissions. This fixes badge saving on Android 13+.

Sharing badges has been slightly improved: Many apps don't read the EXTRA_STREAM property, even if a valid (and readable) URI is provided there. The EXTRA_TEXT property shouldn't contain a description, like was done before, but the actual text content of the file. This is now what badge sharing does, allowing text content of badges to be shared to many more apps. Additionally, the MIME type "text/*" doesn't really exist, so text/plain is used instead, though this could be changed to text/json.

Unfortunately, this change resets all saved cliparts and badges of the user. Since they have previously been saved in external storage, it should be feasible to import at least all the old badges. Once importing cliparts becomes possible, the first issue will be solved as well.

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -12,9 +12,6 @@
 	<uses-feature android:name="android.hardware.bluetooth" android:required="true"/>
     <uses-feature android:name="android.hardware.bluetooth_le" android:required="true"/>
 
-    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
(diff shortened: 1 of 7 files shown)


kleinesfilmroellchen:
Can the PR title validation be run again?

adityastic (approved):
Thanks! Lgtm

PR #968, opened 2024-07-05
fix: BLE won't work on Android 11 and below

### LINKED ISSUE #957: Fix legacy app for Android version 11 and lower
Include Android 11 as a target for the app or lower. This is a follow up to https://github.com/fossasia/badgemagic-android/issues/928

Promote this app to beta testing on the app store.

**1) Manifest**
: If your app targets Android 11 (API level 30) or lower, declare the following permissions in your app's manifest file:

[BLUETOOTH](https://developer.android.com/reference/android/Manifest.permission#BLUETOOTH) is necessary to perform any Bluetooth classic or BLE communication, such as requesting a connection, accepting a connection, and transferring data.
[ACCESS_FINE_LOCATION](https://developer.android.com/reference/android/Manifest.permission#ACCESS_FINE_LOCATION) is necessary because, on Android 11 and lower, a Bluetooth scan could potentially be used to gather information about the location of the user.
photo_6230917659030043862_y

> This is what we need: https://developer.android.com/develop/connectivity/bluetooth/bt-permissions#assert-never-for-location
> 
> Strongly assert that your app doesn't derive physical location If your app doesn't use Bluetooth scan results to derive physical location, you can make a strong assertion that your app never uses the Bluetooth permissions to derive physical location. To do so, complete the following steps:
> 
> Add the android:usesPermissionFlags attribute to your BLUETOOTH_SCAN permission declaration, and set this attribute's value to neverForLocation.
> 
> Note: If you include neverForLocation in your android:usesPermissionFlags, some BLE beacons are filtered from the scan results. If location isn't otherwise needed for your app, remove the ACCESS_FINE_LOCATION permission from your app's manifest.
> 
> The following code snippet shows how to update your app's manifest file:
> 
> ```
> <manifest>
>     
>     <uses-permission android:name="android.permission.BLUETOOTH_SCAN"
>                      android:usesPermissionFlags="neverForLocation" />
> 
>     
>     <strike><uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" /></strike>
>     ...
> </manifest>
> ```

This permission flag I've already added in the last commit.

**2) Update Prominent Disclosure Screen**
Clearly state what permissions are needed.

We can try the following message or suggest a more suitable one:

> Badge Magic uses location data to enable Bluetooth LE and to transfer data to the badges. No location data is transferred to external devices or our servers.

kienvo:
Fixes #957

Changes:
- revert c22c2432 (#936)
- only ask for Location and show Prominent Disclosure on Android 11 and below

Had been tested on Android 10.

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -12,6 +12,9 @@
 	<uses-feature android:name="android.hardware.bluetooth" android:required="true"/>
     <uses-feature android:name="android.hardware.bluetooth_le" android:required="true"/>
 
+    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION"
+        android:maxSdkVersion="30"/>
+
     <application
         android:name=".BadgeMagicApp"
         android:allowBackup="true"
--- a/android/src/main/java/org/fossasia/badgemagic/others/BadgeMagicPermission.kt
+++ b/android/src/main/java/org/fossasia/badgemagic/others/BadgeMagicPermission.kt
@@ -16,11 +16,14 @@ class BadgeMagicPermission private constructor() {
 
     val allPermissions = arrayOf(
         Manifest.permission.WRITE_EXTERNAL_STORAGE,
+        Manifest.permission.ACCESS_FINE_LOCATION,
         Manifest.permission.BLUETOOTH_SCAN,
         Manifest.permission.BLUETOOTH_CONNECT,
     )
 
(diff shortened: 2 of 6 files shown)


kienvo:
The pipeline failed as described at #969

mariobehling:
@kienvo The CI is failing. Could you check, please?

kienvo:
> @kienvo The CI is failing. Could you check, please?

It's failing because of missing keys. I separated a build check only for pull requests at #971.

adityastic (approved):

Jhalakupadhyay (approved):

Issue #954, opened 2024-06-23
Add the UI for displaying the vectors on the screen.


Jhalakupadhyay:
When the user clicks on the prefix icon on the text field, a dialog containing the vectors in tabular form should appear from which the user can select the vector to display on the badge.

PR #972, opened 2024-07-13
feat: Added the UI to display the vectors on prefix Icon Click.

Jhalakupadhyay:
#954 

1. Added the Gridview to display the vectors on prefix Icon click.
2. Display of vectors on the text field when selected by the user.
3. adding a placeholder for the images in the text controller.


lib/constants.dart
```diff
@@ -14,3 +14,46 @@ const String aniRight = 'assets/animations/ic_anim_right.gif';
 const String effFlash = 'assets/effects/ic_effect_flash.gif';
 const String effInvert = 'assets/effects/ic_effect_invert.gif';
 const String effMarque = 'assets/effects/ic_effect_marquee.gif';
+
+//list containing the location of all the assets
+List<String> vectors = [
```
adityastic: Instead of maintaining the list of vectors. Can you try to read all vectors from this location? We don't want to maintain a list as well as the asset where we could infer by reading the AssetManifest.json

lib/bademagic_module/utils/image_utils.dart
```diff
@@ -1,32 +1,40 @@
...
 class ImageUtils {
-  // Define target width and height for the bitmap
-  static double targetHeight = 11.0;
-  static double targetWidth = 44.0;
+  late double originalHeight;
+  late double originalWidth;
 
-  //function to load and scale the svg according to the badge size
-  Future<ui.Image> scaleSVG(String svg) async {
+  late ui.Picture picture;
```
adityastic, line 12: Sharing the picture when not designing the class to handle per image operations would be confusing. If this class is going to process one image anyway. Remove gateways to loadSVG, scaleSVG, etc by making them private

lib/view/splashscreen.dart
```diff
@@ -9,16 +11,23 @@ class SpalshScreen extends StatefulWidget {
...
+  InlineImageProvider cacheImageProvider =
+      GetIt.instance<InlineImageProvider>();
   @override
   void initState() {
-    // TODO: implement initState
     super.initState();
+    _startImageCaching();
+  }
+
+  Future<void> _startImageCaching() async {
```
adityastic: We wanted to remove the splash screen. Do we want to still want to process the images here?

adityastic (commented):
Everything else LGTM! Great job!

adityastic (approved):

Issue #973, opened 2024-07-14
Work on the Logic to transfer the hex of the text and vectors to the badge.


Jhalakupadhyay:
Right now the  vectors can be seen on the grid and on the text field when the user clicks on any one of the vectors but we need to generate the hex for the vectors and text and transfer the data to the badge efficiently.

PR #974, opened 2024-07-17
feat: Added the logic to transfer the text and vectors together on the badge

Jhalakupadhyay:
#973 
Updated the Image util class to transfer the vectors perfectly to the badge.

### DIFF
--- a/lib/bademagic_module/utils/image_utils.dart
+++ b/lib/bademagic_module/utils/image_utils.dart
@@ -1,5 +1,6 @@
 import 'dart:ui' as ui;
 import 'dart:ui';
+import 'package:badgemagic/bademagic_module/utils/byte_array_utils.dart';
 import 'package:badgemagic/bademagic_module/utils/converters.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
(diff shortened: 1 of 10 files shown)


test/data_to_bytearray_converter_test.dart
```diff
@@ -5,11 +5,13 @@ import 'package:badgemagic/bademagic_module/models/speed.dart';
 import 'package:badgemagic/bademagic_module/utils/byte_array_utils.dart';
 import 'package:badgemagic/bademagic_module/utils/converters.dart';
 import 'package:badgemagic/bademagic_module/utils/data_to_bytearray_converter.dart';
+import 'package:badgemagic/providers/getitlocator.dart';
 import 'package:flutter_test/flutter_test.dart';
 
 void main() {
-  DataToByteArrayConverter converter = DataToByteArrayConverter();
+  Future.delayed(const Duration(seconds: 2));
```
adityastic: Is this needed?

lib/providers/imageprovider.dart
```diff
@@ -33,6 +32,9 @@ class InlineImageProvider extends ChangeNotifier {
   //else delte operation is performed.
   int controllerLength = 0;
 
+  //Map to store the hex values of the characters
+  final Map<int, List<String>> hexStrings = {};
```
adityastic: Unused?

lib/bademagic_module/utils/image_utils.dart
```diff
@@ -170,18 +150,31 @@ class ImageUtils {
...
     await _loadSVG(asset);
-    final ui.Image scaledImage = await _scaleSVG(30, 120);
+    ui.Image image =
+        await picture.toImage(originalWidth.toInt(), originalHeight.toInt());
+    final ui.Image scaledImage = await _scaleSVG(image, 30, 120);
     return _trimSVG(scaledImage);
   }
 
   //function to generate the LED hex from the given asset
   Future<List<String>> generateLedHex(String asset) async {
```
adityastic, line 160: Why not start working on bytes rather than Strings?

adityastic: Adds an extra layer of encryption/decryption

test/data_to_bytearray_converter_test.dart
Jhalakupadhyay: removed it was testing something and left it by mistake

lib/providers/imageprovider.dart
Jhalakupadhyay: yes I created this because I was thinking that I should cache the hex values of the vectors that the user clicks maybe that could be helpful.

Jhalakupadhyay: so that when the user selects the same vector again the value will be directly fetched by the map.

adityastic: Your image provider should handle that. Wrap the images inside this class which caches the image object and the cache hex values. Use the provider to cache lazily when the image is used once

lib/bademagic_module/utils/image_utils.dart
Jhalakupadhyay: It is going to be very complicated in dart there is a lot of things that can get way too complicated inorder to do so and maybe it increase the computation time.

Jhalakupadhyay: Maybe this is something that we can work on afterwards if you agree.

lib/providers/imageprovider.dart
Jhalakupadhyay: ok I will update it in the next PR

lib/bademagic_module/utils/image_utils.dart
adityastic: Sure thing, We can always create a great framework around it to manage complexity. Let's park this as an improvement we can look at after we finish the app

lib/providers/imageprovider.dart
adityastic: Sure, can you remove it for now and the delay added

adityastic (approved):

Issue #979, opened 2024-07-25
Add the logic to transfer the user drawn clipart to the badge.


Jhalakupadhyay:
Need to implement a logic that will transfer the user-drawn clipart to the badge.

PR #981, opened 2024-08-04
feat: Added the logic to display the user drawn clipart to the badge.

Jhalakupadhyay:
#979

## Summary by Sourcery

Add logic to display user-drawn clipart on the badge by storing clipart in the image cache with filenames and keys. Enhance image cache handling and update related methods and widgets to support the new functionality. Update tests to reflect changes.

New Features:
- Added logic to display user-drawn clipart on the badge by storing the clipart in the image cache with a filename and key.

Enhancements:
- Modified the image cache to use a map with keys that can be either integers or lists containing filenames and keys.
- Updated the `addToCache` and `_addImageDataToCache` methods to handle filenames for user-drawn clipart.
- Enhanced the `Converters` class to read image data from files and convert it to LED hex format for user-drawn clipart.
- Improved the `InlineImageProvider` to handle insertion of images using keys that can be either integers or lists.
- Refactored the `VectorGridView` widget to use the updated image cache keys.

Tests:
- Updated the `Converters` test to accommodate the new parameter in the `convertBitmapToLEDHex` method.


lib/bademagic_module/bluetooth/base_ble_state.dart
```diff
@@ -45,8 +45,6 @@ abstract class RetryBleState extends BleState {
         attempt++;
         if (attempt < _maxRetries) {
           logger.d("Retrying ($attempt/$_maxRetries)...");
-          await Future.delayed(
```
adityastic, line 48: We should have some delay when retrying. You can reduce the delay but Retrying would not make sense if the failure is transient and you're trying again without waiting

adityastic (approved):

Issue #877, opened 2023-05-11
Update dependencies in the project


mariobehling:
Since the inception of this project some time has passed and some library dependencies are out of date by now. Please update the dependencies of the project and ensure it still builds without errors.

swapniltitoo:
I would like to work on this issue. If open please assign this to me! Thanks.

adityastic:
> I would like to work on this issue. If open please assign this to me! Thanks.

Sure!

AnsahMohammad:
@swapniltitoo  are you working on this ?
Otherwise i would like to work on this

meghanagottapu:
@mariobehling can i work on this?

jaivsh:
Hi @adityastic @mariobehling please assign me this! Thanks!

ANUKULJHA1506:
@adityastic  please assign me this task.

PR #984, opened 2024-08-14
Update dependencies in the project #877

ANUKULJHA1506:
Updated all the dependencies, but the `extended_text_field` dependency needed to be downgraded because it required a higher SDK version than what was supported by the prescribed SDK levels.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  - package-ecosystem: "pub" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 5  
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,28 +5,8 @@
 
 version: 2
 updates:
-  - package-ecosystem: "pub" # See documentation for possible values
-    directory: "/" # Location of package manifests
+  - package-ecosystem: "pub" # Flutter/Dart dependencies
+    directory: "/" # Location of pubspec.yaml
     schedule:
-      interval: "weekly"
-    open-pull-requests-limit: 5  
-name: Auto-merge Dependabot PRs
-
-on:
-  pull_request:
-    types:
-      - labeled
-
-jobs:
-  automerge:
-    if: github.actor == 'dependabot[bot]' && github.event.pull_request.labels.*.name contains 'automerge'
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout code
-        uses: actions/checkout@v2
-
-      - name: Merge PR
-        run: |
-          git checkout ${{ github.head_ref }}
-          git merge --no-ff ${{ github.ref }}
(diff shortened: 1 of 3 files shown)


adityastic (commented):
Interesting. Are we talking about flutter sdk wrt to the downgrade or minimum android compilation sdk? I'm curious to know why 15 worked for us before

adityastic:
@ANUKULJHA1506 It would be much better if you can setup dependabot instead which updates dependencies automatically rather than us periodically updating them

ANUKULJHA1506:
@adityastic  i m  working on dependabot ,And well yes in my android studio it was saying to have higher sdk version to use this as dependency as when i was doing pub get , it was constantly showing errors, thats why i lower the dependency version.

adityastic:
@ANUKULJHA1506 lets close this pr and create the changes for dependabot and then we can update them granularly

ANUKULJHA1506:
@adityastic ok , is their any slack or any other platform where i can connect with the fossasia community members.

adityastic:
@ANUKULJHA1506 indeed, there is a gitter channel in the README. Also can you remove pubspec changes. We'll get dependabot to make them. Thanks!

.github/dependabot.yml
```diff
@@ -0,0 +1,32 @@
...
+version: 2
+updates:
+  - package-ecosystem: "pub" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 5  
+name: Auto-merge Dependabot PRs
+
+on:
```
adityastic: Not sure why this is here. This is not dependabots config

adityastic (changes requested):

.github/dependabot.yml
ANUKULJHA1506: these will send automatically PRs to update the versions , upto 5 PRs can be send at a time, then you can check them and push it in main branch.

ANUKULJHA1506: do you want me to make some changes in depentabot config cause it is working fine in my forked repo.

adityastic: There is a job configuration which is not relevant to dependabot.yml

adityastic: I dont think it belongs here

pubspec.lock
```diff
@@ -167,10 +167,10 @@ packages:
     dependency: "direct dev"
     description:
       name: flutter_lints
-      sha256: "9e8c3858111da373efc5aa341de011d9bd23e2c5c5e0c62bccf32438e192d7b1"
+      sha256: "3f41d009ba7172d5ff9be5f6e6e6abb4300e263aab8866d2a0842ed2a70f8f0c"
       url: "https://pub.dev"
     source: hosted
-    version: "3.0.2"
+    version: "4.0.0"
```
adityastic, line 173: We don't need changes in pubspec* files as dependabot will create PR's to update them. Can you remove both changes

adityastic (changes requested):

pubspec.lock
ANUKULJHA1506: ok , I'm looking into it.

ANUKULJHA1506:
@adityastic is dependabot.yml need some changes or it is fine i have removed the  two jobs that u asked for, now dependabot will create only PR's to update them.

adityastic:
@ANUKULJHA1506 both the pubspec file are still present

ANUKULJHA1506:
@adityastic are you asking about dependabot.yml file or any other file ?

adityastic:
@ANUKULJHA1506 please read the last comment

PR #1012, opened 2024-09-22
feat: Added the updated base structure for handling the animations

Jhalakupadhyay:
## Summary by Sourcery

Add new utility functions for byte and binary data conversion, introduce a new provider for managing badge animations and drawing, and refactor existing components to integrate with the new provider structure. This update enhances the animation handling capabilities and improves the modularity of the badge rendering system.

New Features:
- Introduce a new utility function to convert byte arrays to binary arrays and binary strings to 2D lists, enhancing data manipulation capabilities.
- Add a new `badgeAnimation` function in the `Converters` class to handle badge animations based on message input.
- Implement a new `DrawBadgeProvider` class to manage badge drawing and animation states, replacing the previous `DrawBadgeProvider` implementation.

Enhancements:
- Refactor the `BadgeWidget` to accept a grid parameter, allowing for more flexible badge rendering.
- Update the `HomeScreen` to initialize and manage animations using the new `DrawBadgeProvider`.

Chores:
- Replace the `drawbadge_provider.dart` with `badgeview_provider.dart` across multiple files to align with the new provider structure.


lib/bademagic_module/utils/byte_array_utils.dart
```diff
@@ -28,3 +28,56 @@ List<int> hexStringToByteArray(String hexString) {
   logger.d(data.length);
   return data;
 }
+
+List<List<int>> byteArrayToBinaryArray(List<int> byteArray) {
+  List<List<int>> binaryArray = List.generate(11, (_) => []);
```
adityastic: Use badge height

lib/bademagic_module/utils/converters.dart
```diff
@@ -43,6 +43,20 @@ class Converters {
...
   }
 
+  void badgeAnimation(String message) async {
+    if (message == "") {
+      //geerate a 2d list with all values as 0
+      List<List<int>> image =
+          List.generate(11, (i) => List.generate(44, (j) => 0));
+      badgeList.setNewGrid(image);
+    } else {
+      List<String> hexStrings = await messageTohex(message);
```
adityastic, line 53: We should avoid these 3 operations and decide a scheme to store messages to one format which can be used across the app. List<List<bool>>  would be a good candidate. It does not make sense to have these translations for every step when we could built our logic just on one kind of data

lib/providers/badgeview_provider.dart
```diff
@@ -0,0 +1,119 @@
...
+  int counter = 0;
+  Timer? timer;
+  //List that contains the state of each cell of the badge for home view
+  List<List<bool>> homeViewGrid =
+      List.generate(11, (i) => List.generate(44, (j) => false));
+
+  //List that contains the state of each cell of the badge for draw view
+  List<List<bool>> drawViewGrid =
+      List.generate(11, (i) => List.generate(44, (j) => false));
+
```
adityastic, line 18: This needs to be cleaned. 2 seperate widgets for draw and display seperately

adityastic (changes requested):
Nice comments from sourcery to look at

lib/providers/badgeview_provider.dart
JhalakUpadhyay7: this I will clear in seperate PR where I will add separate custom paint class for the drawbadge

lib/bademagic_module/utils/converters.dart
JhalakUpadhyay7: But in order to get the bitmap (List<List<int>>) we would first need to have the hexstring and then convert it to bitmap.
Can you clarify what you mean by avoiding the operations.

lib/bademagic_module/utils/byte_array_utils.dart
JhalakUpadhyay7: ok

lib/bademagic_module/utils/converters.dart
adityastic: Hex String -> Byte Array, Byte Array -> Binary Array, Binary Array -> Bool Array. Why not Hex String -> Bool Array

JhalakUpadhyay7: I have not seen this conversion so far is it possible?

JhalakUpadhyay7: Give me some time I will look into it and if I can make it then it will be very efficient.

adityastic: So we're trying to minimize the number of conversions. HexString to something you can work off of everywhere. Maybe 2 conversions is fine but not more

adityastic: So HexString -> ByteArray is fine. Then maybe ByteArray -> Bool Array when needed.

adityastic: I don't see a point in holding up 4 bytes of memory whereas you can get away with just a bit

adityastic: Correction, flutter uses 8 bytes int so more memory usage for no reason

adityastic (approved):

PR #1144, opened 2025-01-05
chore: configure and build Dev Container for flutter development

ZauberNerd:
This commit adds a Dev Container configuration to allow building and testing the Android Flutter application without additional setup on the host.
It also presents a method of using the Dev Container in a cloud workspace (GitHub Codespaces) while still being able to push the builds to a connected Android device.

## Summary by Sourcery

Build:
- Add a Dev Container configuration with necessary dependencies and tools for Flutter and Android development.


adityastic (approved):
Would make development really convenient! Thanks for looking after dev xp

PR #1193, opened 2025-02-15
added font support to text

nope3472:
This PR adds support for selecting fonts in the Home Screen UI, addressing the issue raised by @mariobehling. Proprietary apps already support fonts, and now this feature is implemented here as well.
WhatsApp Image 2025-02-15 at 15 39 31

## Summary by Sourcery

Add support for custom fonts, font sizes, and text color selection in the Home Screen UI. Update the badge animation to reflect the selected font settings.

New Features:
- Add font selection support to the Home Screen UI, allowing users to choose from a list of Google Fonts and customize the font size and color of the displayed text on the badge.

Build:
- Add `flutter_colorpicker` package dependency.
- Add `google_fonts` package dependency.

### DIFF
--- a/lib/providers/animation_badge_provider.dart
+++ b/lib/providers/animation_badge_provider.dart
@@ -43,7 +43,7 @@ class AnimationBadgeProvider extends ChangeNotifier {
   int _animationSpeed = aniSpeedStrategy(0);
   Timer? _timer;
 
-  //List that contains the state of each cell of the badge for home view
+  // List that contains the state of each cell of the badge for home view
   List<List<bool>> _paintGrid =
       List.generate(11, (i) => List.generate(44, (j) => false));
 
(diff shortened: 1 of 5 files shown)


adityastic:
@nope3472 I don't think he meant font selection for app fonts. He meant more font options for badge itself, ie the data we transfer to badges

mariobehling:
Yes, the font should show up on the badge preview and then the same should be transfered to the badge itself.

See below, please.

413533279-ad39deac-366c-4c27-97f9-2220dd9e8ba3

nope3472:
@mariobehling so i should keep the font style selector  and implement that same style to the animation

mariobehling:
Yes, please. If possible could you integrate the font style selector into the text field directly using the dropdown on the right?

The font should be changed in the animation, but it should also be transmitted in that way and show up like that on the badge itself.

PR #1195, opened 2025-02-15
fix: update github events to be avoid negation

adityastic:
## Summary by Sourcery

CI:
- Update iOS, Android screenshot and iOS build workflows to only run on push events instead of on every event except for pull requests


AsCress (approved):
@adityastic Good to go :))

PR #1221, opened 2025-03-05
chore: overwrite screenshots

AsCress:
_Screenshots_ on the _App Store_ aren't getting updated when we make a new release (see https://apps.apple.com/us/app/badge-magic/id6740176888).
Possibly, we can configure _fastlane_ to overwrite screenshots whenever we _deliver_ the builds to production ?
@adityastic I observed that you had commented out this line. Was there any specific reason for that ?

## Summary by Sourcery

Chores:
- Enable overwriting screenshots when delivering builds to production.


adityastic:
That's indeed because: https://github.com/fastlane/fastlane/pull/22036/files. Fastlane has not updated the new dimensions for screenshots and will fail to verify if we upload screenshots. You might want to have a look at our options

AsCress:
@adityastic Thanks for the information !

AsCress:
@adityastic However, the PR you tagged aims at updating dimensions for 11' screenshots. Does the same problem exist with 13' ones too, since we use those ones?

adityastic:
> @adityastic However, the PR you tagged aims at updating dimensions for 11' screenshots. Does the same problem exist with 13' ones too, since we use those ones?

That's correct, even with the new 6.9' screenshots.

AsCress:
@adityastic Ohh is it ? Is _fastlane_ not in active development ? But in any case, let's explore other options for us maybe.

adityastic:
@AsCress indeed. Might want to explore what our options are

PR #1259, opened 2025-03-26
feat: Added Support for Fonts

### LINKED ISSUE #239: Implement Option to Change Font
Currently there is only one standard font implemented. Please implement an option to change the font.

### LINKED ISSUE #1134: Add support for fonts
Currently proprietary apps already support fonts. Please also add the support here.

The font should show up on the badge preview and then the same should be transferred to the badge itself.

See draft below, please.

413533279-ad39deac-366c-4c27-97f9-2220dd9e8ba3

Vishveshwara:
Fixes #239 
Fixes #1134

## Changes 
- Created a Drop Down Menu with 7 google Fonts ('Roboto','Open Sans','Lato', 'Poppins','Montserrat','Orbitron', 'Lexend')
- Optimized for characters exceeding the 11x8 grid by combining two characters into an 11x16 layout.
- Dynamically adjusted height to ensure that characters with descenders, such as 'p' and 'q', are fully visible on the screen.
- Enabled effortless font expansion by simply adding new entries to the FontProvider.
- Guaranteed full compatibility with existing emojis while preserving the default font setting for a consistent user experience.

## Screenshots / Recordings  

https://github.com/user-attachments/assets/59240dea-53e3-46a6-b267-309a8d4de304

**Checklist**: 
- [x] **No end of file edits**: No modifications done at end of resource files.

## Summary by Sourcery

Add support for custom fonts in the badge display, allowing users to select from a predefined list of Google Fonts and dynamically render text with improved character rendering

New Features:
- Implemented a dropdown menu to select from 7 different Google Fonts
- Added dynamic font rendering that supports characters with descenders
- Created a flexible font system that can easily expand with new font options

Enhancements:
- Optimized text rendering for LED matrix display
- Implemented intelligent character scaling for wide characters
- Added dynamic height adjustment for characters with descenders


samruddhi-Rahegaonkar:
@Vishveshwara Thanks for working !
just few changes the selected fonts should also display on the textfield area. and also noticed one thing there are inconsistent pixel distances between the characters which we don't want. after each character their should have only the distance of a 1 pixel. Please go through the code once and check it.

Vishveshwara:
@samruddhi-Rahegaonkar The issue with spacing of characters is due to the width of characters like "w","h","m" which does not fit in 11x8 grid which is the size we are using for the default font. so for now , I have written a code in which a wide character can occupy 11x16 grid and hence the character does not get cut off but the character occupies double the space.

https://github.com/user-attachments/assets/b36bf056-7264-405f-8b77-5f4e192a88d8

the Solutions for this issue: 
1) Using a smaller font (But causes poor visibility of text. Currently I am using font size 16, font size 14 seems to be fine but still has the spacing issue and anything lower is a problem)
2) use 11x16 grid for all the characters (but the space between characters is very big, can be good for very wide fonts).
3) using a variable grid dimension based on the character width (Best possible solution, but would need a different pipeline from how we are processing the text now)


I will try to work on this, @Jhalakupadhyay can you guide me on which possible solution would be the best

samruddhi-Rahegaonkar:
@Vishveshwara That’s a good temporary fix, but it introduces an inconsistency where some characters take double the space, breaking uniform alignment.

samruddhi-Rahegaonkar:
@Vishveshwara you might try 
Scale down wide characters to fit within 11x8 without cutting off.

Jhalakupadhyay:
@Vishveshwara nice going but there are few changes needed.
1. as suggested by @samruddhi-Rahegaonkar the font should also change in the textfield.
2. Try to maintain a cache for the characters already render such that we don't need to process the characters whose hex is already generated once this will increase the speed of our solution 
3. Let's try scalling down the characters to fit into 11x8 grid and see how it looks I think it will be good option.

I will look closely to PR once and then can tell you better, but we can go for the above changes first.

Vishveshwara:
https://github.com/user-attachments/assets/dc1c95ae-23b5-41df-ac77-5a65499b2c67

Challenges
- scaling down the characters to fit into 11x8 grid , makes the characters unreadable so I had to go with the variable column solution

@Jhalakupadhyay @adityastic please review this PR.

Changes
1. Added Consistent spacing between characters by using variable columns for each character( for example 'i' uses 3 columns and 'W' uses 12 columns).
2. Added Caching for all the fonts currently, and added a feature to generate cache in settings screen for newer fonts.
3. Cache is currently stored as a 2d Binary Bool array.
4. Font changes in the text field.

adityastic:
Why do we have 13k loc changed?

adityastic (changes requested):
I don't think we need 13k lines to have font support in our app.

Vishveshwara:
> I don't think we need 13k lines to have font support in our app.

> @Vishveshwara nice going but there are few changes needed.
> 
> 1. as suggested by @samruddhi-Rahegaonkar the font should also change in the textfield.
> 2. Try to maintain a cache for the characters already render such that we don't need to process the characters whose hex is already generated once this will increase the speed of our solution
> 3. Let's try scalling down the characters to fit into 11x8 grid and see how it looks I think it will be good option.
> 
> I will look closely to PR once and then can tell you better, but we can go for the above changes first.

@adityastic
The 13k lines are a result of caching font characters as 2d bool matrices in **_lib/badgemagic_module/utils/font_cache.dart_**. As suggested by Jhalak, I implemented a caching mechanism to avoid regenerating character matrices repeatedly.

adityastic:
> > I don't think we need 13k lines to have font support in our app.
> 
> > @Vishveshwara nice going but there are few changes needed.
> > 
> > 1. as suggested by @samruddhi-Rahegaonkar the font should also change in the textfield.
> > 2. Try to maintain a cache for the characters already render such that we don't need to process the characters whose hex is already generated once this will increase the speed of our solution
> > 3. Let's try scalling down the characters to fit into 11x8 grid and see how it looks I think it will be good option.
> > 
> > I will look closely to PR once and then can tell you better, but we can go for the above changes first.
> 
> @adityastic
> The 13k lines are a result of caching font characters as 2d bool matrices in **_lib/badgemagic_module/utils/font_cache.dart_**. As suggested by Jhalak, I implemented a caching mechanism to avoid regenerating character matrices repeatedly.

I don't think he meant caching it directly in the repository 😅

Vishveshwara:
> I don't think he meant caching it directly in the repository 😅

@adityastic 

I'm sorry for the confusion. I noticed that the default font was being cached in "lib/badgemagic_module/utils/data_to_bytearray_converter.dart", and so I assumed it was acceptable to cache them directly. However, I understand that this might not be the ideal approach. The code can work without caching as well, so I would appreciate your guidance on where the best place to cache these characters would be.

adityastic:
> > I don't think he meant caching it directly in the repository 😅
> 
> @adityastic 
> 
> I'm sorry for the confusion. I noticed that the default font was being cached in "lib/badgemagic_module/utils/data_to_bytearray_converter.dart", and so I assumed it was acceptable to cache them directly. However, I understand that this might not be the ideal approach. The code can work without caching as well, so I would appreciate your guidance on where the best place to cache these characters would be.

We should cache every character in memory while conversation is being performed and reuse the character or the subset if we see repeated entries to improve execution time. That's all to it I guess. Let me know if you have any other thoughts.

Vishveshwara:
> > > I don't think he meant caching it directly in the repository 😅
> > 
> > 
> > @adityastic
> > I'm sorry for the confusion. I noticed that the default font was being cached in "lib/badgemagic_module/utils/data_to_bytearray_converter.dart", and so I assumed it was acceptable to cache them directly. However, I understand that this might not be the ideal approach. The code can work without caching as well, so I would appreciate your guidance on where the best place to cache these characters would be.
> 
> We should cache every character in memory while conversation is being performed and reuse the character or the subset if we see repeated entries to improve execution time. That's all to it I guess. Let me know if you have any other thoughts.

@adityastic 
I've implemented the character caching as you suggested. Please review it, whenever you are available.

Vishveshwara:
@adityastic can you please review this pr? I have done the changes you have mentioned.

Vishveshwara:
@nope3472 i have read your scrum update, as you are working on the font change feature , you can check this PR out and mention your comments.

samruddhi-Rahegaonkar:
Hello, @Vishveshwara Have you tested this implementation on badge ?

Vishveshwara:
> Hello, @Vishveshwara Have you tested this implementation on badge ?

Not yet , I will test it today and update you

samruddhi-Rahegaonkar:
> > Hello, @Vishveshwara Have you tested this implementation on badge ?
> 
> Not yet , I will test it today and update you

yup! It will be great to review further!

Vishveshwara:
@samruddhi-Rahegaonkar there seems to be some issue in the segmenting or padding logic i have done, the first part of the text is working properly , and even in the virtual badge i didnt see these artifacts. I will try to fix it and ping you when i am done.

https://github.com/user-attachments/assets/0c77e8dd-a1cb-49ca-b88c-3ba91cc1259d

font_test_app

Vishveshwara:
WhatsApp Image 2025-06-30 at 16 52 00_1a0d8dea

https://github.com/user-attachments/assets/103bc72a-b424-4540-9c3f-bd44b6ef256f

Vishveshwara:
@samruddhi-Rahegaonkar its working on the hardware now, you can pull this code and test it out.

lib/bademagic_module/utils/converters.dart
samruddhi-Rahegaonkar, line 1: renderTextToMatrix() calls toImage() and byte analysis for every character heavy for long strings to optimize it you can Render entire string at once, Split based on change in descender requirement, so try to optimize code as much as you can.

samruddhi-Rahegaonkar, line 1: Try to optimize _processCustomFontMessage and _processDefaultFont into a single function.

samruddhi-Rahegaonkar (changes requested):

lib/bademagic_module/utils/converters.dart
Vishveshwara: This shouldn't matter as if a font is not selected , _processDefaultFont is called which uses the method we originally have to process the text(using `converter.charCodes[char]`) just segmenting is added to handle bitmaps and texts.

Vishveshwara: I have optimized renderTextToMatrix() Converting the pixel data to boolean matrix code.

For each character toImage() is called just once , and the cache will be saved for each character, we can't cache the entire string. The current method is the best way to do it , Please go through the code line by line and understand it.

and the image size is very less 11x16 at max , this is very fast to compute.

samruddhi-Rahegaonkar (approved):
LGTM! Nice work.

hpdang:
@nope3472 please review this

Vishveshwara:
@Jhalakupadhyay can you please review this PR

hpdang (approved):

Jhalakupadhyay:
@Vishveshwara Looks nice have you tested it thoroughly with all the fonts and all teh characters ? If yes I think we can go ahead with this PR.

hpdang:
@Vishveshwara could you please resolve conflicts so we can merge this?

nope3472 (approved):

samruddhi-Rahegaonkar:
Hey, @Vishveshwara Will you please resolve conflicts once.

mariobehling (approved):

PR #1325, opened 2025-06-13
fix: Incorrect animation screen options

### LINKED ISSUE #1248: Behaviour change in the "Snowflake" animation
Hei!

In the "old" app, the "Snowflake" animation had a really neat behavior which seems to have been removed in the new version. This is easiest to see by setting the animation speed to max, and doing the following steps:

1. Add enough text so that it "overflows" (meaning it requires more than one display cycle to show it all
2. Make sure that the text contains at least a few spaces at the "right side" of the first screen, so that there are no characters that are rendered partly.
3. Enable "snowflake"

e.g. input

`test- test`

With the old app, the text would render like this (one "screenful" per line)

```
testX
test
```

With the new app, it renders like this:

```
testX
 test
```

In the old app, this would create the illusion that the letter `X` was "blinking", since the second rendering always made sure the content started a the leftmost pixel of the screen.

In the new app, the content isn't moved to the left any more, so the illusion breaks - the whole text bounces left and right, instead of just one character appearing and disappearing.

I hope this makes sense....

Could we get the old behavior back? 😁

### LINKED ISSUE #1370: Animation screen options incorrect
The animation screen does not show the correct gif animations and implementations for
* animation
* snowflake
* picture

Expected: Compare the features with the proprietary legacy app and implement the correct functionality in the badge preview on the app and the badge. Ensure the preview and actual badge show the same.

samruddhi-Rahegaonkar:
Fixes #1248 
Fixes https://github.com/fossasia/badgemagic-app/issues/1370

## Changes 
-Updated Animation according to proprietary app.

## Screenshots / Recordings  

**Checklist**: 
- [x] **No end of file edits**: No modifications done at end of resource files.

## Summary by Sourcery

Improve Bluetooth handling in `BadgeMessageProvider` by automating enablement on Android, adding platform-specific prompts, timeouts, and error logging, and streamline the data transfer flow; refactor `SnowFlakeAnimation` to simplify frame computation and prevent rendering glitches by aligning animation slices and bounding checks.

Enhancements:
- Automate Bluetooth enabling on Android with error handling and timeouts, and add platform-specific prompts for iOS
- Restructure `BadgeMessageProvider` to unify adapter state checks and streamline data generation and transfer
- Refactor snowflake animation to compute frame count once, align frames to the left, and guard against out-of-bounds grid accesses


nope3472 (changes requested):
The frame calculation now uses animationIndex % framesCount instead of animationIndex ~/ badgeWidth % framesCount, which ignores badgeWidth and skews the animation timing/sequencing.

samruddhi-Rahegaonkar:
Uploading Screen Recording 2025-06-20 at 11.00.38 AM.mov…

samruddhi-Rahegaonkar:
@adityastic @Jhalakupadhyay Please review this too.

samruddhi-Rahegaonkar:
@nope3472 Will you please review it i will add recording of output?

nope3472 (commented):
The function processAnimation lacks input validation for the size and structure of processGrid and canvas. This may cause runtime errors if the input arrays do not match the expected dimensions.

hpdang:
@samruddhi-Rahegaonkar what is the status here?

samruddhi-Rahegaonkar:
@hpdang This feature has been completed from my side also added recordings regarding it and updated the codes according to the review.

samruddhi-Rahegaonkar:
@nope3472 Review the Latest changes!

mariobehling (changes requested):
@samruddhi-Rahegaonkar Please follow the reviews and conversation in the chat channel 

So the correct one should be:
    ANIMATION = 0x05,
    SNOWFLAKE = 0x06,
    PICTURE = 0x07

mariobehling:
fixed-animation

Screenshot from 2025-07-09 19-11-14

samruddhi-Rahegaonkar:
@mariobehling According to the proprietary app following are the recordings for animation 
1.Animation
2.Picture
3.Snowflake


https://github.com/user-attachments/assets/bbb5e2e4-f3ce-4ee5-80e1-e2144ced8ed0


https://github.com/user-attachments/assets/5470f5ab-666c-4439-babc-f8ecc128a705


https://github.com/user-attachments/assets/68d83a04-c112-4c3d-af0a-9245a4f2ca87

hpdang:
@samruddhi-Rahegaonkar please share an APK with the changes in your PR so I can test it?

hpdang:
@samruddhi-Rahegaonkar the Animation effect works as expected on the physical LED badge. However, the in-app preview (what users see on their phone screen before sending it to the badge) does not accurately reflect the actual animation that plays on the badge. This is confusing for users, because they expect the preview to show exactly what will appear on the badge. Please test it again and fix it.

samruddhi-Rahegaonkar:
@hpdang we could only show in app preview if user set three messages in textfield and saved as animation. if we saved each word as individual animation then in saved badge preview it will show still image because we cannot create animation with one word.

samruddhi-Rahegaonkar:
@hpdang in the Homescreen preview they see what animation looks like.

hpdang:
@samruddhi-Rahegaonkar users should be able to understand how their text will appear on the badge without having to save anything first.

If the user’s input is just a single word that fits within one screen, the animation remains fixed on that screen, and the preview should reflect this behavior.

**In short:** the in-app preview must always accurately represent what will actually happen on the badge, regardless of the input length or animation type.

Maybe you can try: 
 
- Test with multiple text lengths 1) Very short text (fewer characters than the badge can display in one screen), 2) Text that exactly fits one full screen, 3) Longer text that requires scrolling across multiple screens. 
- Detect when the text fits on a single screen and show a static animation in the preview.
- Ensure the preview mimics the badge output frame by frame.

hpdang:
@samruddhi-Rahegaonkar the preview doesnt not correctly reflect what is on the physical badge. 
@nope3472 please review this PR, and share your idea how this should be done


https://github.com/user-attachments/assets/ce831bbd-f28f-4916-bf09-6aa3c19b91a3


https://github.com/user-attachments/assets/e98390e4-e520-4473-96fc-8d34b240840e

hpdang:
@samruddhi-Rahegaonkar what is the status here?

samruddhi-Rahegaonkar:
@hpdang this is the most technically correct solution that we can write for this issue. isn't it @nope3472  ?

samruddhi-Rahegaonkar:
@hpdang The badge and preview use completely separate rendering systems, and unless both are driven by the exact same data stream and logic, 100% visual sync in full-width word-switching is not feasible.
Even if the preview and badge share the same data stream, they still won’t match in full-width word-switching, because they interpret and animate that data differently.
The badge reads the byte array directly and animates based on its own firmware logic.
The badge does not use our animationIndex or frame slicing. So even after giving the same processGrid, the frame 2 is not necessarily badge's frame 2.

samruddhi-Rahegaonkar:
@nope3472 I have done with this feature. Please review and if there are any changes suggest the changes or please aprove it. so i can move to another issue.

samruddhi-Rahegaonkar:
@nope3472

nope3472 (changes requested):
-There is repeated logic for mapping animation indices to modes and vice versa across multiple files (providers/animation_badge_provider.dart, providers/badge_message_provider.dart, providers/saved_badge_provider.dart).
-Suggestion: Create a central utility/helper for animation/mode/index mapping. This makes future changes less error-prone and the codebase easier to maintain.


APART FROM THIS THE LOGIC IS TECHNICALLY CORRECT.

samruddhi-Rahegaonkar:
@nope3472 I think all have there seperate working functionality making single utility will be hard to manage isn't it ?

AnimationBadgeProvider handles grid updates, animation rendering, timer lifecycle.
BadgeMessageProvider handles BLE data packaging and transmission.
SavedBadgeProvider handles saved state logic and restoring animations from JSON.

samruddhi-Rahegaonkar:
@nope3472

samruddhi-Rahegaonkar:
@nope3472 Please Aprove the changes as discussed.

nope3472 (approved):

hpdang:
@mariobehling please approve this

hpdang (approved):

mariobehling (approved):

PR #1539, opened 2025-12-08
fix(win32): remove WS_VISIBLE and add assert in GetClientArea

### LINKED ISSUE #1528: Win32Window::window_handle_ may be nullptr immediately after CreateWindow
In Win32Window::Create, the window_handle_ member is set only in WndProc when handling WM_NCCREATE:

  auto that = static_cast<Win32Window*>(window_struct->lpCreateParams);
  that->window_handle_ = window;

This creates a race condition because if any code calls Show(), SetChildContent(), GetClientArea(), or Destroy() immediately after CreateWindow returns, window_handle_ may still be nullptr. This can lead to crashes or undefined behavior.

Proposed fix:
- Assign window_handle_ immediately after CreateWindow returns, or
- Ensure that any code using window_handle_ waits until WM_NCCREATE is processed.

g-k-s-03:
closes #1528 
Description:
Previously, window_handle_ was set only in WndProc during WM_NCCREATE. This caused a race condition if methods like Show(), SetChildContent(), or GetClientArea() were called immediately after CreateWindow(), potentially leading to crashes.

Changes:
Set window_handle_ immediately after CreateWindow() in Win32Window::Create().
Added a safeguard in WndProc to ensure window_handle_ is set.
Updated dark mode handling to check Windows version before applying attributes.

Impact:
Calls to window methods are now safe immediately after creation.
Prevents crashes due to uninitialized window handle.

Testing:
Verified Show(), SetChildContent(), GetClientArea(), and Destroy() immediately after creation.
Dark mode behavior tested on supported Windows builds.

## Summary by Sourcery

Resolve race conditions around Win32 window handle initialization and make dark mode handling conditional on OS support.

Bug Fixes:
- Initialize the Win32 window handle immediately after window creation and guard usages to prevent race-condition crashes when calling window methods early.
- Avoid calling window APIs like Show(), SetChildContent(), and GetClientArea() when the window handle is not yet available.
- Prevent use of immersive dark mode attributes on Windows versions that do not support them.

Enhancements:
- Show and update the Win32 window immediately upon creation instead of relying solely on later Show() calls.

### DIFF
--- a/windows/runner/win32_window.cpp
+++ b/windows/runner/win32_window.cpp
@@ -53,6 +53,27 @@ void EnableFullDpiSupportIfAvailable(HWND hwnd) {
   FreeLibrary(user32_module);
 }
 
+// Checks if the current Windows version supports dark mode
+bool IsDarkModeSupported() {
+  // DWMWA_USE_IMMERSIVE_DARK_MODE requires Windows 10 Build 17763 (version 1809) or higher
+  OSVERSIONINFOEXW osvi = {};
+  DWORDLONG const dwlConditionMask = VerSetConditionMask(
+      VerSetConditionMask(
+          VerSetConditionMask(
+              0, VER_MAJORVERSION, VER_GREATER_EQUAL),
+          VER_MINORVERSION, VER_GREATER_EQUAL),
+      VER_BUILDNUMBER, VER_GREATER_EQUAL);
+  
+  osvi.dwOSVersionInfoSize = sizeof(osvi);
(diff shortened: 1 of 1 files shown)


g-k-s-03:
I’m closing this PR as the changes here are not directly relevant to the current GSoC focus on Ell-ena/Devr.AI. 
I appreciate the review and feedback received, and I’m happy to revisit or reopen this PR if it becomes relevant in the future.

Thank you!

PR #1381, opened 2025-07-23
feat: created animations that can be transferred to the badge

nope3472:
Fixes #641

## Changes

* Added seamless, classic Pacman animation with mouth and block destruction effect, including proper reset and state handling for smooth looping and transfer to the badge.
* Unified animation selection and transfer logic for all custom and standard animations, improving maintainability and user experience.
* Automatically switch to text animation mode when typing text, preserving text-based animations’ original speed mapping.
* Updated the animation selection UI to display all animations in rows of three, using relevant icons for custom animations without assets.
* Fixed animation state persistence and ensured all custom animations start fresh when reselected.
* Improved the Pacman animation transfer so the last frame is always clean, preventing visual glitches at the end of the animation on the badge.

## Screenshots / Recordings

> N/A

**Checklist**:

* [ ] **No hard coding**: I have used resources from `constants.dart` without hard coding any value.
* [ ] **No end of file edits**: No modifications done at end of resource files.
* [ ] **Code reformatting**: I have reformatted code and fixed indentation in every file included in this pull request.
* [ ] **Code analyzation**: My code passes analyses run in *flutter analyze* and tests run in *flutter test*.

## Summary by Sourcery

Introduce four new custom badge animations (Pacman, Chevron, Diamond, Broken Hearts) with seamless looping, state reset, and dedicated transfer routines, unify animation selection and transfer logic, update the animation UI to display custom animations with icons in a 3-column grid, and improve text animation handling by auto-switching modes on typing.

New Features:
- Add Pacman, Chevron, Diamond, and Broken Hearts custom animations with dedicated badge transfer functions
- Extend AnimationBadgeProvider.handleAnimationTransfer to route transfers to custom animation routines
- Update the animation tab UI to display custom animations with relevant icons arranged in rows of three

Bug Fixes:
- Fix Pacman end-of-animation glitch by adding a clean final frame and ensuring proper state reset

Enhancements:
- Unify animation selection and transfer logic across all custom and standard animations
- Automatically switch to text animation mode when typing to preserve original speed mapping
- Reset animation state on reselection for smooth looping

PR #1614, opened 2026-03-14
fix: preserve user's animation mode for single-badge transfers (#1506)

### LINKED ISSUE #1506: First badge is transfered with wrong transition.
When I try to transfer multiple badges to my device, the first one has always the "Animation" transition and not the one the app shows. Every other badge transfers correct. 

I checked also the exported json for the badge and there is the right value inside. So I think it's something during transfer.

Note: This happens in the Beta Version.

aditya27b:
When transferring a single saved badge, the first badge was always forced to Mode.animation (0x05) regardless of the user's saved mode. This caused wrong transitions for fixed/laser/custom animation badges.

The override was originally intended for multi-slot badge transfers (PR #1381) to create seamless continuous display cycling. It now only applies when transferring 2+ badges to fill multiple slots.

- Single-badge transfers now use correct saved mode
- Multi-slot seamless marquee behavior preserved
- Backward compatible with all existing saved badge JSON files

Fixes #1506

## Changes 
lib/providers/badge_message_provider.dart:222

Changed from:
if (isSavedBadge && data.messages.isNotEmpty) {
  mode: Mode.animation, // Forced on all
To:
if (isSavedBadge && data.messages.length > 1) {
  mode: Mode.animation, // Only for multi-slot

## Screenshots / Recordings  
N/A

**Checklist**: 
- [ ] **No end of file edits**: No modifications done at end of resource files.

only info no error/warning
image
All tests passed
image

## Summary by Sourcery

Adjust badge transfer behavior to only force animation mode when transferring multiple saved messages, preserving the user’s chosen mode for single-badge transfers.

Bug Fixes:
- Preserve the saved display mode when transferring a single saved badge instead of always forcing animation mode.
- Restrict the seamless marquee animation override to multi-slot badge transfers with more than one message.

### DIFF
--- a/lib/providers/badge_message_provider.dart
+++ b/lib/providers/badge_message_provider.dart
@@ -210,14 +210,14 @@ class BadgeMessageProvider {
     Data data;
     if (jsonData != null) {
       data = fileHelper.jsonToData(jsonData);
-      if (isSavedBadge && data.messages.isNotEmpty) {
+      if (isSavedBadge && data.messages.length > 1) {
         final old = data.messages[0];
         final newMessage = Message(
           text: old.text, // use the already-padded hex string
           flash: old.flash,
           marquee: old.marquee,
           speed: old.speed,
-          mode: Mode.animation, // Force seamless marquee
+          mode: Mode.animation, // Force seamless marquee for multi-slot badges
         );
         data = Data(messages: [newMessage, ...data.messages.skip(1)]);
       }


aditya27b:
Note on Sourcery's assessment: The issue title mentions 'first badge transferred with wrong transition' but the actual report is about single-badge transfers from SaveBadgeCard. The multi-slot override (Mode.animation on first slot) is intentional per PR #1381 for seamless cycling — this fix preserves that behavior while only fixing the single-badge case. 
The ❌ on Sourcery's assessment is because it read the fix as 'multi-slot first badge still overridden' which is correct by design.

ctrlVnt (commented):
Hi @aditya27b 
Thank you very much for your PR🙏

Did you already see #1739 ?
Actually I deleted all `ìf` block because when badge was saved didn't transfer the good transition but forced "animation" transition, what do you think about?

ctrlVnt:
Thank you to your PR!

problem was solved with https://github.com/fossasia/badgemagic-app/pull/1754

PR #1669, opened 2026-04-28
fix: add supportedLocales and enable language switching for new translations

DELTA-45-G:
Depends on #1653

- Added supportedLocales for new languages (el, ml, my, ta, te)
- Updated language selector dropdown
- Enabled dynamic locale switching using appLocale

Added a sample Tamil translation ("settings") to verify functionality.
Other languages currently fall back to English until translations are added via Weblate.
WhatsApp Image 2026-04-28 at 11 23 14 PM

## Summary by Sourcery

Add new language options and enable dynamic locale switching in the app settings.

New Features:
- Expose Greek, Malayalam, Tamil, Telugu, and Burmese as selectable languages in the settings screen.
- Register the new locales as supported app locales for runtime selection.

Enhancements:
- Switch settings screen localization to use the generated AppLocalizations context instead of the custom localization service.


DELTA-45-G:
Thanks for the feedback!

Updated the dropdown language labels to use AppLocalizations instead of hardcoded strings.
Added the corresponding keys in app_en.arb and regenerated localization files.
Language switching is working correctly. Currently, only keys available in the ARB files are translated, and missing keys fall back to English as expected.

ctrlVnt (approved):
LGTM

EDIT : I found some issues, see my comments bellow

lib/main.dart
```diff
@@ -90,6 +90,11 @@ class MyApp extends StatelessWidget {
                 Locale('en'),
                 Locale('hi'),
                 Locale('it'),
+                Locale('el'),//Greek
```
ctrlVnt: @DELTA-45-G can you just delete the comments please?

ctrlVnt (changes requested):
@DELTA-45-G Good job

Can you also delete this string from settings_screen.dart please ?

``` dart
final List<String> languages = ['en', 'hi', 'it'];
```

lib/main.dart
DELTA-45-G: Done. I removed the unused language list and locale comments, rebased on the latest branch updates, and pushed the changes. Could you please review again?

Saksham-Sirohi (commented):
Copilot flagged the following:

No issues in the sampled diff.

ctrlVnt (changes requested):
Hi @DELTA-45-G 

To merge this PR we need also the translations for the added languages, otherwise it can create confusion to see "Greek" but no changes (because arb file is empy)

To merge this PR we need translations before.

I think it's better to add just one language and the ARB file with the corresponding translation

ctrlVnt:
Hi @DELTA-45-G 
I close this PR due to my last comment, but you can open a new one with one new language if you want :)

PR #1700, opened 2026-05-18
fix: large integer issue for flutter web compatibility

Monika-18777:
## Problem

The app failed to run properly on Flutter Web because the large integer value exceeded JavaScript's safe integer limit.

## Solution

Replaced the large integer value with a web-safe integer in `special_text_field.dart`.

## Impact

* Fixes Flutter Web compatibility issue
* Prevents runtime integer overflow problems on web
* Keeps behavior consistent across platforms

## Summary by Sourcery

Replace an oversized inline image start index with a JavaScript-safe integer to restore Flutter Web compatibility and prevent overflow issues.

### DIFF
--- a/lib/view/special_text_field.dart
+++ b/lib/view/special_text_field.dart
@@ -48,6 +48,10 @@ class InlineImage extends SpecialText {
   }
 }
 
+// Sentinel start index for InlineImage runs.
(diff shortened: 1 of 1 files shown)


rahul31124 (approved):
LGTM!

marcnause (approved):

PR #1725, opened 2026-05-28
fix: preserve clipart transparency

### LINKED ISSUE #1726: Cliparts are saved with white background instead of transparency
## Description

Saved cliparts currently get a white background instead of a transparent one.

This is happening because the empty pixels in `convert2DListToUint8List` are being saved as white pixels instead of transparent pixels.

The issue is not clearly visible inside the app since the cliparts are shown on white backgrounds, but it becomes noticeable on dark or colored backgrounds where a white box appears around the clipart.

## Expected Behavior
Cliparts should be saved with a transparent background so only the drawing is visible.

> Current behaviour recording

https://github.com/user-attachments/assets/197f738a-f650-436b-bec4-4f1ed552b8d3

bhavjsh:
Fixes #1726
## Changes
- Updated `convert2DListToUint8List` in `lib/bademagic_module/utils/image_utils.dart`
- Changed "off" pixels from opaque white to fully transparent
- Kept "on" pixels unchanged as solid black

## Testing
- Verified locally by changing the clipart background color to red
- Confirmed that only the black clipart is rendered and the background remains transparent
- Reverted the temporary red background test change before committing

## Recording  

https://github.com/user-attachments/assets/cf52cfbc-7b68-489c-97ca-cf89d2319a83

## Summary by Sourcery

Bug Fixes:
- Fix clipart rendering so that background pixels are fully transparent while foreground pixels remain solid black.


rahul31124 (approved):
Looks good. Previously, the cliparts were saved with a white background. Now they will be saved with a transparent background, which is better

Issue #1728, opened 2026-05-28
Preset Clipart Inconsistency


bhavjsh:
## Issue

Some preset cliparts appear differently in the preview bar compared to how they look inside the clipart picker grid.

The selected clipart should render consistently across both the picker and preview areas, but currently certain shapes/icons look altered or mismatched after selection.

<img width="591" height="1280" alt="Image" src="https://github.com/user-attachments/assets/6ccca2fc-14d7-4963-a055-3480f9f46a71" />
<img width="591" height="1280" alt="Image" src="https://github.com/user-attachments/assets/d87dc75a-baad-4a2d-baa7-3503133111ca" />

suryansh00001:
I'd like to take this issue

bhavjsh:
> I'd like to take this issue

Go ahead : )

suryansh00001:
@bhavjsh Check this out

PR #1740, opened 2026-06-04
fix: clipart grouping and sizing

### LINKED ISSUE #1727: Fix inconsistent clipart sizing and improve clipart grouping
## Problem

The clipart section currently feels unorganized and inconsistent.

- Similar cliparts are scattered in different places instead of appearing together.
- Some cliparts look very small while others appear too large.

This makes browsing cliparts harder and gives an uneven UI experience.

## Expected Behavior

- Similar cliparts should appear together in nearby positions or same rows.
- All cliparts should have a more consistent visual size.
- The clipart grid should look cleaner and easier to browse.

> Current clipart sizing issue

https://github.com/user-attachments/assets/999f0fda-b358-415d-9f39-858791b8c379

> Random order of cliparts in picker grid 

Image

bhavjsh:
Fixes  #1727

## Changes 
- Fixed clipart sizing so cliparts render at a consistent balanced size both in the clipart picker grid and in the badge/preview bar (the actual LED output), instead of some looking tiny, oversized, or cropped.
- Grouped the built-in cliparts by category in the picker (faces, arrows, hearts, shapes, etc.) so similar icons sit together and are easier to find.
- Made all 8 directional arrows the same size by fixing their SVG viewBox, so they no longer float in empty space or get clipped at the bottom.
-  Hid two duplicate cliparts from the picker the left arrow (which actually rendered as a right arrow) and the duplicate thumbs-up while keeping their assets so existing saved badges aren't affected.

> Recording

https://github.com/user-attachments/assets/5083f3a2-62fd-4fde-83b4-dd76e41fc426

> Grouping similar items 

photo_2026-06-04_17-20-16


rahul31124 (approved):
@bhavjsh, Nice work! LGTM!

ctrlVnt (changes requested):
Most of the icons seem improved except these ones, which on the contrary have lost their charm:
Image
Image
Image
Image

ctrlVnt (changes requested):
Now it's better, there is still this one that I found worse than before:

Image

Also I didn't find this one in the list:

Image

Ps. this change improved also some others, as bottom arrow!

bhavjsh:
@ctrlVnt  I intentionally hid the "thumbs up" and "right side arrow" cliparts, as they were duplicates and I've changed the "Mail" clipart image. This could be fully resolved by addressing #1728 (since this PR is focused on making clipart sizing as consistent as possible), but for now I've made these changes:
mail_clipart

mariobehling (approved):

Jhalakupadhyay:
@bhavjsh removing the scale function wont help because if new cliparts that are not according to the badge dimension are added that makes the clipart rendering broken.

bhavjsh:
> @bhavjsh removing the scale function wont help because if new cliparts that are not according to the badge dimension are added that makes the clipart rendering broken.

Hello @Jhalakupadhyay I'm not removing the scaling, it already fits every clipart to 11 rows no matter the size so the new ones are fine, it's just that the picker now reuses that same bitmap so it matches the preview instead of using a separate render

Jhalakupadhyay:
> > @bhavjsh removing the scale function wont help because if new cliparts that are not according to the badge dimension are added that makes the clipart rendering broken.
> 
> Hello @Jhalakupadhyay I'm not removing the scaling, it already fits every clipart to 11 rows no matter the size so the new ones are fine, it's just that the picker now reuses that same bitmap so it matches the preview instead of using a separate render

Just saw that you have renamed the function which was not necessary but the logic is fine. Please try to keep the changes minimum only what is required makes PR easier to review.

PR #1795, opened 2026-07-03
fix: fixed preview same as the badge result

ctrlVnt:
Partial fixes #1775

## Changes 
I tried to switch of one column the transfer but I didn't manage to do it. I realized that fixed preview was different from result, so I decided to adapt the preview. Even if it doesn't fix completely the problem now user knows what to expect.

## Screenshots / Recordings  
Screenshot_20260703-134329
Screenshot_20260703-134342

**Checklist**: 
- [x] **No hard coding**: I have used resources from `constants.dart` without hard coding any value.
- [x] **No end of file edits**: No modifications done at end of resource files.
- [x] **Code reformatting**: I have reformatted code and fixed indentation in every file included in this pull request.
- [x] **Code analyzation**: My code passes analyzations run in _flutter analyze_ and tests run in _flutter test_.

PR #1785, opened 2026-06-25
fix: imported badge editability and custom clipart portability

xar43:
Fixes #1769 
# Fix imported badge editability and custom clipart portability

## Overview

While working on this issue, I found that badge exports only stored the rendered badge payload (`messages`). This was sufficient for displaying the badge, but it did not preserve enough information to fully reconstruct the original badge after import.

As a result:

* Imported badges could not always be edited reliably.
* The original user input was lost after export/import.
* Personalized cliparts were tied to the device on which they were created.
* Importing the same badge on another device could result in missing symbols.
* Legacy badge formats required special handling and were not fully compatible with the current import flow.

This PR updates the badge import/export pipeline to preserve the original badge state while maintaining compatibility with existing badge files and the current rendering workflow.

---

### Added `originalText`

A new optional field, `originalText`, has been added to the badge JSON structure.

This stores the exact text entered by the user before rendering. By preserving the original input, imported badges can now be reopened and edited without relying on filename-based reconstruction or fallback placeholders.

### Added `customCliparts`

A new optional field, `customCliparts`, has been added to store personalized clipart data.

The field contains the raw matrix representation of custom cliparts that are actually used within a badge, allowing them to be transferred together with the badge export.

---

## Export Improvements

The badge save flow has been updated to preserve additional information during export.

### Original Text Preservation

When a badge is saved, the original user input is now passed through the save pipeline and embedded into the exported JSON.

This ensures that the editing experience can be restored later, even after transferring the file to another device.

### Custom Clipart Detection

The export process now scans the original text for clipart references.

For personalized cliparts (`data_*`):

* The clipart is located in the local cache.
* Its matrix data is extracted.
* The matrix is embedded into the exported badge JSON.

Built-in cliparts are not duplicated, which helps keep exported files smaller while still preserving custom content.

---

### Restore Original Badge Content

During import, the stored `originalText` is restored and saved through `BadgeTextStorage`.

This allows the editor to reopen imported badges with the same text and symbols originally used by the creator.

### Automatic Custom Clipart Installation

When an imported badge contains embedded custom cliparts:

* The import flow checks whether each clipart already exists locally.
* Missing cliparts are recreated automatically.
* Recreated cliparts are saved to disk.
* The cliparts are registered in the rendering cache.

This allows custom symbols to work immediately after import without requiring any manual setup.

### Cross-Device Badge Portability

Because custom cliparts are now bundled with the badge export itself, badges can be transferred between devices without losing personalized content.

---

## Legacy Format Support

[...]

PR #1802, opened 2026-07-06
fix: resolve text clipping and animation lookup issue #1796

### LINKED ISSUE #1796: [Discussion] Fixed animation is not fixed
Fixed animation in the app doesn't move at all, but if i transfer the result is the same effect of "slitting" animation
The app:

## The app

https://github.com/user-attachments/assets/bb303e62-5ae3-4b01-ab93-5571b1720474

## The badge

https://github.com/user-attachments/assets/be3e5b8a-7aa7-4350-8c4d-b0da3f6357e4

xar43:
Closes #1796

## Fix "Fixed" Animation Bug

This PR fixes the issue where the **Fixed** animation doesn't behave as expected. Instead of remaining fixed, it can clip the text and may even get transferred as the **Splitting** animation.

### What was causing the issue?

The app was identifying the selected animation by comparing object instances (`==`). This works only when the exact same instance is reused. If a new instance of the same animation class is created, the comparison fails, causing the app to fall back to the wrong animation during transfer.

### What's changed?

I updated `lib/badge_animation/animation_abstract.dart` to make animation comparisons based on the animation type instead of object identity.

By overriding `==` and `hashCode` in the `BadgeAnimation` base class to compare `runtimeType`, different instances of the same animation (for example, two `FixedAnimation` objects) are now treated as equal.

This ensures the correct animation is recognized and transferred, preventing the unexpected fallback behavior.

https://github.com/user-attachments/assets/a52ec0b7-ae5e-45a1-a55b-994562cc5ee8

## Summary by Sourcery

Adjust badge animation handling to correctly identify and apply fixed animations without clipping or unintended fallbacks.

Bug Fixes:
- Ensure fixed badge animations are identified by type rather than instance to prevent incorrect animation transfers.
- Prevent fixed animation text clipping by clamping the horizontal offset to non-negative values.
- Avoid incorrectly defaulting to an animation index when no current animation is found, allowing proper handling of the idle state.
- Stop badge animations cleanly when no message and no special animation are selected by cancelling timers and resetting the animation index.

### DIFF
--- a/lib/badge_animation/ani_fixed.dart
+++ b/lib/badge_animation/ani_fixed.dart
@@ -1,3 +1,4 @@
+import 'dart:math';
 import 'package:badgemagic/badge_animation/animation_abstract.dart';
 
 class FixedAnimation extends BadgeAnimation {
@@ -5,7 +6,7 @@ class FixedAnimation extends BadgeAnimation {
   void processAnimation(int badgeHeight, int badgeWidth, int animationIndex,
       List<List<bool>> processGrid, List<List<bool>> canvas) {
     int newWidth = processGrid[0].length;
-    int horizontalOffset = (badgeWidth - newWidth) ~/ 2;
+    int horizontalOffset = max(0, (badgeWidth - newWidth) ~/ 2);
 
     for (int i = 0; i < badgeHeight; i++) {
       for (int j = 0; j < badgeWidth; j++) {
--- a/lib/badge_animation/animation_abstract.dart
+++ b/lib/badge_animation/animation_abstract.dart
@@ -1,4 +1,10 @@
 abstract class BadgeAnimation {
   void processAnimation(int badgeHeight, int badgeWidth, int animationIndex,
       List<List<bool>> processGrid, List<List<bool>> canvas);
+
+  @override
+  bool operator ==(Object other) => runtimeType == other.runtimeType;
+
+  @override
+  int get hashCode => runtimeType.hashCode;
 }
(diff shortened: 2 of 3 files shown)


ctrlVnt:
Hi @xar43, I think this is a duplicate of #1795

xar43:
Hi @ctrlVnt ,

Thank you for your feedback! I completely understand why it might look like a duplicate at first glance since both address the "Fixed animation" behavior, but I would like to explain the critical architectural differences:

1. **Root Cause Analysis vs. Layout Tweak:** As I originally pointed out in my comment on this issue 3 days ago, the core problem of text clipping stems from a negative `horizontalOffset` in `ani_fixed.dart` when text length exceeds the badge width. PR #1795 attempts to stretch the preview bounds loop to align with the clip, whereas this PR (#1802) corrects the underlying math using `max(0, ...)` so text naturally aligns to the left edge and gracefully overflows.
2. **The Hardware Transfer Bug (Crucial Fix):** PR #1795 does not address why the animation incorrectly changes to "Splitting" during actual hardware transfer. This PR solves that root cause by overriding `==` and `hashCode` based on `runtimeType` inside the `BadgeAnimation` abstract class, ensuring the provider never fails the lookup or silently falls back during transfer.
3. **State Management:** This PR also safely preserves the user's selected animation state when the text field is temporarily cleared, rather than forcing a hard-reset to `LeftAnimation`.

Since this PR completely resolves both the visual clipping math, state management, and the hardware transfer glitch (which was the core of the issue reported here), it offers a permanent, comprehensive fix. 

I would be extremely grateful if you could review this architectural approach. Thank you so much for your time and guidance! 

pls look this one also  #1785

ctrlVnt:
Hi @xar43,
Do you have a badge for testing?

ctrlVnt:
@xar43 I close this issue because it create confusion. So sorry was my wrong, I marked that issue as bug but this need a discussion because this animation was implemented before that I start contributing and I need to understand the exact scope

Anyway thank you for your time!

PR #1812, opened 2026-07-13
feat: Badge renaming

ctrlVnt:
Fixes #1359

PR parent #1808

## Changes 
- fixed bug for search configured names, now we find correctly listed names only
- add renaming the badge button when we are connected to the badge
- when we connect to the badge name is added in the list automatically, even if we rename the badge
- fixed bug duplicated names in badge list

> [!IMPORTANT]
> You need a badge with next-generation protocol implemented. I tested with this version and **it works to me**: https://github.com/fossasia/badgemagic-firmware/actions/runs/27161014451

## Screenshots / Recordings  

https://github.com/user-attachments/assets/2f77181c-c8da-4a27-93a9-6d14a8b9ca8c

**Checklist**: 
- [x] **No hard coding**: I have used resources from `constants.dart` without hard coding any value.
- [x] **No end of file edits**: No modifications done at end of resource files.
- [x] **Code reformatting**: I have reformatted code and fixed indentation in every file included in this pull request.
- [x] **Code analyzation**: My code passes analyzations run in _flutter analyze_ and tests run in _flutter test_.

PR #1828, opened 2026-07-25
fix: Discover OEM badges missing 0xFEE0 in adverts

### LINKED ISSUE #1827: OEM (LSLED) badges are never discovered: scan filters on 0xFEE0, which is not advertised
## Description

Badges running the **OEM firmware** (BLE name `LSLED`) are never discovered by the app. Scanning always ends with the `Device not found.` toast after the 15 s timeout, even though the badge is powered on, in range, and works correctly with the stock vendor app.

Badges flashed with the open `badgemagic-firmware` are discovered normally, which is why this is easy to miss during development.

## Steps to reproduce

1. Take a badge running unmodified OEM firmware (advertises as `LSLED`).
2. Open the app, leave **Scan mode** at `any`.
3. Trigger a transfer.

**Expected:** `Device found. Connecting...`
**Actual:** `Device not found.` after 15 s.

## Root cause

The OEM firmware **does not include the `0xFEE0` service UUID in its advertisement packet** — the service only becomes visible after connecting. Discovery is gated on that UUID in two places, and both reject the badge:

`lib/bademagic_module/bluetooth/scan_state.dart`

```dart
final matchesUuid = device.services.contains(targetServiceUuid);   // line 44
...
await UniversalBle.startScan(
  scanFilter: ScanFilter(
    withServices: [targetServiceUuid],                             // line 81
  ),
);
```

`withServices` is pushed down to the native scanner, so the device is filtered out before it ever reaches Dart. And `device.services` on a scan result holds the *advertised* service list, not the GATT table, so the callback check would drop it even if a result got through.

## Evidence

A `bleak` scan finds the badge and, after connecting, shows `fee0`/`fee1` present in its GATT table — confirming the service exists but is simply absent from the advertisement:

```
Connecting to 9C:7F:64:AE:77:BE (LSLED) ...
  service fee0
    char  fee1  [read,write,notify]
```

Decompiling the stock vendor app (`Bluetooth LED Name Badge` 3.5.1) shows it never filters on the service UUID. It scans with an **empty** filter and matches on the device name instead:

```java
// com/yannis/ledcard/ble/BLEScanner.java
this.mScanFilter = new ScanFilter.Builder().build();      // no filter
arrayList.add(new ScanFilter.Builder().build());
this.mScanner.startScan(arrayList, this.mScanSettings, this.mScanCallback);

// onBatchScanResults
String name = scanResult.getDevice().getName();
if (!TextUtils.isEmpty(name) && name.contains(BleDevice.DEVICE_NAME)) { ... }

// com/yannis/ledcard/ble/BleDevice.java
public static final String DEVICE_NAME = "LS";
public static final UUID serviceUUID        = 0000fee0-0000-1000-8000-00805f9b34fb;
public static final UUID characteristicsUUID = 0000fee1-0000-1000-8000-00805f9b34fb;
```

The write path uses exactly the same `fee0`/`fee1` UUIDs as this app, so **discovery is the only difference** between the working vendor app and this one.

## Suggested fix

Supplying `withNamePrefix` alongside `withServices` is enough, because of how `universal_ble` treats it. From `UniversalBlePlugin.kt`:

```kotlin
val usesCustomFilters = filter?.usesCustomFilters() ?: false   // true iff withNamePrefix is set
if (usesCustomFilters) {
    universalBleFilterUtil.scanFilter = filter                 // match in the plugin
} else {
    scanFilters = filter?.toScanFilters(filterServices)        // native UUID filter
}
safeScanner.startScan(scanFilters, settings, scanCallback)
```

Setting `withNamePrefix` switches the plugin to custom-filter mode, where the list handed to the **native** scanner is empty (an unfiltered scan, as the vendor app does) and matching moves into the plugin, where the filter kinds are OR'd — `UniversalBleFilterUtil.kt:37`, mirrored in Dart at `universal_ble_filter_util.dart:61`.

Pairing that with a name check OR'd against the existing UUID check in the scan callback fixes OEM discovery while leaving open firmware — which does advertise `0xFEE0`, and whose name is user-changeable — matching exactly as before.

## Environment

- App built from `development` @ 42c98bc
- `universal_ble` 2.1.0
- Badge: CH582M, 11x44, OEM firmware, advertising as `LSLED`

bkero:
Fixes #1827

## Changes

Badges running the **OEM firmware** (BLE name `LSLED`) were never discovered — scanning always ended in `Device not found.` after the 15 s timeout, even with the badge powered on, in range, and working correctly with the stock vendor app. Badges flashed with the open `badgemagic-firmware` were unaffected, which makes this easy to miss during development.

### Root cause

The OEM firmware does **not** include `0xFEE0` in its *advertisement packet* — the service only becomes visible after connecting. Discovery was gated on that UUID in two independent places, and both rejected the badge:

- `startScan(scanFilter: ScanFilter(withServices: [serviceUuid]))` — this is pushed down to the **native** scanner, so the device was filtered out before it ever reached Dart.
- `device.services.contains(targetServiceUuid)` in the scan callback — `device.services` on a scan result holds the *advertised* service list, not the GATT table, so it would have dropped the device even if a result had got through.

### Fix

- Added `badgeNamePrefixes` to `lib/globals/globals.dart`, next to the existing `serviceUuid` / `characteristicUuid` constants so the BLE identifiers stay together.
- Passed `withNamePrefix: badgeNamePrefixes` to `ScanFilter`. This is what makes the change small: setting it switches `universal_ble` into custom-filter mode, where the list handed to the native scanner is **empty** (an unfiltered scan) and matching moves into the plugin, with the filter kinds OR'd together.

  ```kotlin
  // UniversalBlePlugin.kt
  val usesCustomFilters = filter?.usesCustomFilters() ?: false   // true iff withNamePrefix is set
  if (usesCustomFilters) {
      universalBleFilterUtil.scanFilter = filter
  } else {
      scanFilters = filter?.toScanFilters(filterServices)
  }
  safeScanner.startScan(scanFilters, settings, scanCallback)
  ```

  OR semantics confirmed at `UniversalBleFilterUtil.kt:37` and mirrored in Dart at `universal_ble_filter_util.dart:61`.
- Changed the callback gate from `matchesUuid && matchesName` to `(matchesUuid || matchesBadgeName) && matchesName`.

Open firmware still matches on `0xFEE0` exactly as before, so nothing regresses for flashed badges, and the separate `BadgeScanMode.specific` name gate is left untouched.

### Reference implementation

The stock vendor app (`Bluetooth LED Name Badge` 3.5.1) scans with an **empty** `ScanFilter` and matches on the name instead:

```java
// com/yannis/ledcard/ble/BLEScanner.java
this.mScanFilter = new ScanFilter.Builder().build();
...
if (!TextUtils.isEmpty(name) && name.contains(BleDevice.DEVICE_NAME)) { ... }

// com/yannis/ledcard/ble/BleDevice.java
public static final String DEVICE_NAME = "LS";
```

Its `fee0`/`fee1` write UUIDs are identical to this app's, so discovery was the only difference. Full details in #1827.

## Screenshots / Recordings

> N/A — no UI change.

## Testing

- `flutter analyze` on both changed files: **No issues found!**
- `dart format --set-exit-if-changed`: clean, 0 files changed
- `flutter test`: **All tests passed!** (22 tests)
- Root cause independently confirmed by a `bleak` scan of the badge (`fee0`/`fee1` present in the GATT table, absent from the advertisement) and by decompiling the vendor app

Note for reviewers: the change is verified by static analysis and against the vendor app's behaviour, but I have not yet been able to confirm end-to-end discovery on a physical OEM badge with this build installed. Independent confirmation from anyone holding an `LSLED` badge would be welcome.

**Checklist**:
- [x] **No hard coding**: the new name prefixes are a named constant in `globals.dart`, placed with the existing BLE UUID constants rather than inlined at the call site.
- [x] **No end of file edits**: no resource files modified.
- [x] **Code reformatting**: `dart format` reports no changes needed.
- [x] **Code analyzation**: `flutter analyze` passes on both files.

## Summary by Sourcery

Relax badge discovery to match OEM firmware badges that do not advertise the FEE0 service UUID while preserving behaviour for open firmware badges.

Bug Fixes:
- Ensure OEM LSLED badges are discoverable by scanning based on name prefixes when FEE0 is absent from advertisements.

Enhancements:
- Centralize badge advertised-name prefixes alongside existing BLE UUID globals for consistent identification.
- Broaden BLE scan filtering to combine service-UUID and name-prefix matching so both OEM and open firmware badges are handled uniformly.

### DIFF
--- a/lib/bademagic_module/bluetooth/scan_state.dart
+++ b/lib/bademagic_module/bluetooth/scan_state.dart
@@ -41,13 +41,19 @@ class ScanState extends NormalBleState {
                 .where((e) => e.isNotEmpty)
                 .toList();
 
+            // device.services here is what was in the advertisement, not the
+            // full GATT table. OEM badges advertise no service UUIDs at all,
+            // so this alone rejects them - match the advertised name too.
             final matchesUuid = device.services.contains(targetServiceUuid);
+            final rawName = (device.name ?? "").trim();
+            final matchesBadgeName =
+                badgeNamePrefixes.any((prefix) => rawName.startsWith(prefix));
 
-            final deviceName = (device.name ?? "").trim().toLowerCase();
+            final deviceName = rawName.toLowerCase();
             final matchesName = mode == BadgeScanMode.any ||
                 normalizedAllowedNames.contains(deviceName);
 
-            if (matchesUuid && matchesName) {
+            if ((matchesUuid || matchesBadgeName) && matchesName) {
               isCompleted = true;
               timeoutTimer?.cancel();
               await UniversalBle.stopScan();
@@ -76,9 +82,15 @@ class ScanState extends NormalBleState {
         },
       );
 
+      // Supplying withNamePrefix switches universal_ble into custom-filter
+      // mode, where the *native* scanner runs unfiltered and matching happens
+      // in the plugin with OR semantics across filter kinds. That is what lets
+      // OEM badges - which advertise no service UUIDs - be seen at all, and it
+      // mirrors what the stock vendor app does.
       await UniversalBle.startScan(
         scanFilter: ScanFilter(
           withServices: [targetServiceUuid],
+          withNamePrefix: badgeNamePrefixes,
         ),
       );
 
--- a/lib/bademagic_module/bluetooth/scan_state.dart
+++ b/lib/bademagic_module/bluetooth/scan_state.dart
@@ -19,6 +19,32 @@ class ScanState extends NormalBleState {
     required this.allowedNames,
   });
 
+  /// Whether [device] is a badge this scan should connect to.
+  ///
+  /// Two independent questions, deliberately kept apart:
+  ///
+  /// 1. Does it look like a badge at all? Satisfied by the advertised service
+  ///    UUID **or** the advertised name. These are OR'd because OEM firmware
+  ///    advertises no service UUIDs, while the open firmware's name is
+  ///    user-changeable - requiring both would exclude one or the other.
+  /// 2. Does it pass the user's badge-selection setting? Only consulted in
+  ///    [BadgeScanMode.specific], and matched on the full name, lower-cased.
+  bool _isTargetBadge(BleDevice device) {
+    // device.services holds what the advertisement carried, not the full GATT
+    // table, so OEM badges contribute nothing here and rely on the name.
+    final looksLikeBadge = device.services.contains(targetServiceUuid) ||
+        matchesBadgeName(device.name);
+
+    if (!looksLikeBadge) return false;
+    if (mode == BadgeScanMode.any) return true;
+
+    final deviceName = (device.name ?? '').trim().toLowerCase();
+    return allowedNames
+        .map((e) => e.trim().toLowerCase())
+        .where((e) => e.isNotEmpty)
+        .contains(deviceName);
+  }
+
   @override
   Future<BleState?> processState() async {
     manager.clearConnectedDevice();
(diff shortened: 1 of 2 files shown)


Saksham-Sirohi (commented):
Copilot flagged the following:

1. OEM prefix matching / unfiltered scan looks right for badges missing 0xFEE0.
2. Will conflict with #1808/#1812 on `scan_state.dart`.
3. Still needs a check on real `LSLED` hardware.

ctrlVnt (changes requested):
Hi @bkero,

thank you for your contribution!
It's an interesting point but I tested with OEM firmware and the app does not find the badge if I set to connect only to LSLED and VBLAB, did you tested on a badge?

Also did you already see the PR #1812?
I fixed a bug where the app couldn't find any name before

PR #1910, opened 2026-08-24
fix: Preserve saved badge transfer mode

### LINKED ISSUE #1863: Bug: Highspeed scrolling on saved badges batch-upload
Hi!
Sorry if the title is a little bit confusing, I wasn't sure how to word it better. Even describing the bug is quite complicated, but I try:

I created several badges consisting of multiple words, with "default" font and "left" transition with speed 8. When sending one of them to the badge, everything is fine. But when I open the "saved badges" page and upload more than one, the first one always speeds up to The Flash like high-speed.

I made a little video to help visualize the case: https://photos.app.goo.gl/1NEHJJbhmZa83bjr6

Here, I made two badges "First test badge" and "Second test badge", both share the same settings (Font default, speed 8, transition left). Uploaded single, both work, uploaded together as batch only the second one works as expected.

This happens on my LSLED provided by fossasia at the last CCC congress in combination with the current beta version of the app (installed today, unfortunately the app has no version information included).

atirna:
## What does this PR do?

Preserves the mode and speed stored in saved badge messages during batch transfer. Previously the transfer path rewrote the first saved message to `Mode.animation`; it now sends decoded saved data unchanged.

## Changes

- remove the saved-badge first-message rewrite from `BadgeMessageProvider`
- pass the decoded saved data to `DataTransferManager` unchanged

Fixes #1863

## Verification

The transfer-path check confirms a saved two-message batch keeps left mode and speed 8. The unit test has been removed as requested; the real-device check remains the final validation.

## Summary by Sourcery

Preserve saved badge transfer settings by keeping decoded message data unchanged.

Bug Fixes:
- Preserve the mode and speed stored in saved badge messages during batch transfers.

Enhancements:
- Send decoded saved badge data through the transfer path without modifying its first message.


ctrlVnt (commented):
It's ok to me the edit to `badge_message_provider.dart` file but I'm not sure about the test.

I'd delete the unit test for the moment, even if can be useful but need to be tested on real device. Even it create some workflow issue

What do you think @rahul31124 and @bhavjsh

ctrlVnt (approved):
Nice job !

rahul31124 (approved):

